PR #10, opened 2021-11-16
Makefile: split build and run

pierreprinetti:
With this patch, the Makefile targets for building and running the tests
are now separate. A conditional is also set on the build target to only
rebuild in case of changes to the OpenStack tests.

Also with this patch, the compiled binary is now called
`openstack-tests` instead of `openstack-test`, as [recommended][1].

The legacy `openstack-test` target is added for backwards compatibility.

[1]: https://github.com/openshift/release/pull/22556#discussion_r735627024

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,4 @@
 *.out
 
 /_output
-/openstack-test
+/openstack-tests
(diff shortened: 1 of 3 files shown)


mandre (commented):
/lgtm

mandre:
Can you make sure to update the README with this new target?

pierreprinetti:
Good point! thanks. done. PTAL

EmilienM:
/lgtm

thanks!

PR #11, opened 2021-12-09
Check masters placement with strict anti affinity

itzikb-redhat:
Check that the masters are placed on different hosts in case of strict
anti affinity

### DIFF
--- a/test/extended/openstack/servergroup.go
+++ b/test/extended/openstack/servergroup.go
@@ -6,42 +6,56 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/servergroups"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+	exutil "github.com/openshift/origin/test/extended/util"
+	yaml "gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 )
 
 var _ = g.Describe("[sig-installer][Feature:openstack] The OpenStack platform", func() {
 	defer g.GinkgoRecover()
 
-	// OCP 4.5: https://issues.redhat.com/browse/OSASINFRA-1300
-	g.It("creates Control plane nodes in a server group", func() {
-		ctx := context.TODO()
+	var computeClient *gophercloud.ServiceClient
+	var ctx context.Context
+	var dc dynamic.Interface
+	var oc *exutil.CLI
+	var masterInstanceUUIDs []interface{}
+	var nodeList *corev1.NodeList
+	var ms dynamic.NamespaceableResourceInterface
+	var serverGroupName string
+
+	oc = exutil.NewCLI("openstack")
+
+	g.BeforeEach(func() {
+		ctx = context.TODO()
 
 		g.By("preparing a dynamic client")
 		cfg, err := e2e.LoadConfig()
 		o.Expect(err).NotTo(o.HaveOccurred())
-		dc, err := dynamic.NewForConfig(cfg)
+		dc, err = dynamic.NewForConfig(cfg)
 		o.Expect(err).NotTo(o.HaveOccurred())
 
-		g.By("getting the IDs of the Control plane instances and their Server group")
-		masterInstanceUUIDs := make([]interface{}, 0, 3)
-		var serverGroupName string
+		g.By("getting the IDs of the Control plane instances")
 		{
-
+			masterInstanceUUIDs = make([]interface{}, 0, 3)
 			clientSet, err := e2e.LoadClientset()
 			o.Expect(err).NotTo(o.HaveOccurred())
 
-			nodeList, err := clientSet.CoreV1().Nodes().List(ctx, metav1.ListOptions{
+			nodeList, err = clientSet.CoreV1().Nodes().List(ctx, metav1.ListOptions{
 				LabelSelector: "node-role.kubernetes.io/master",
 			})
 			o.Expect(err).NotTo(o.HaveOccurred())
 
-			ms := dc.Resource(schema.GroupVersionResource{
+			ms = dc.Resource(schema.GroupVersionResource{
 				Group:    "machine.openshift.io",
 				Resource: "machines",
 				Version:  "v1beta1",
@@ -50,6 +64,17 @@ var _ = g.Describe("[sig-installer][Feature:openstack] The OpenStack platform", 
 				uuid := strings.TrimPrefix(item.Spec.ProviderID, "openstack:///")
 				masterInstanceUUIDs = append(masterInstanceUUIDs, uuid)
 
+			}
+			computeClient, err = client(serviceCompute)
+			o.Expect(err).NotTo(o.HaveOccurred())
+		}
+	})
+
+	// OCP 4.5: https://issues.redhat.com/browse/OSASINFRA-1300
+	g.It("creates Control plane nodes in a server group", func() {
+		g.By("Getting the the Control instances' Server groups")
+		{
+			for _, item := range nodeList.Items {
 				machineAnnotation := strings.SplitN(item.Annotations["machine.openshift.io/machine"], "/", 2)
 				o.Expect(machineAnnotation).To(o.HaveLen(2))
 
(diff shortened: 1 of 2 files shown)


test/extended/openstack/servergroup.go
```diff
@@ -79,7 +80,23 @@ var _ = g.Describe("[sig-installer][Feature:openstack] The OpenStack platform",
...
+			if len(serverGroup.Policies) == 1 {
+				//host_ids := make([]string, 0)
+				host_ids := make(map[string]int)
+
+				for _, server_id := range masterInstanceUUIDs {
+					server, err := servers.Get(computeClient, server_id.(string)).Extract()
+					o.Expect(err).NotTo(o.HaveOccurred())
+					host_ids[server.HostID] += 1
+				}
+				if serverGroup.Policies[0] == "strict-anti-affinity" {
```
pierreprinetti, line 93: as far as I'm aware the "strict" anti-affinity policy is really called `anti-affinity` in OpenStack

pierreprinetti (commented):
The Go logic looks good to me.

Do you mind adding this check to a separate test? That way, we cleanly separate concerns and test results are easier to read.

I am thinking about something like (just a suggestion!):

```Go
g.It("creates Control plane nodes on separate hosts when serverGroupPolicy is anti-affinity", func() {
    if serverGroupPolicy != "anti-affinity" {
        e2eskipper.Skipf("This test only applies when serverGroupPolicy is set to anti-affinity")
    }
    // [...]
```

If you want to avoid code duplication, consider adding "uuid gathering" to [a `g.BeforeEach` stanza](https://onsi.github.io/ginkgo/#extracting-common-setup-beforeeach) (might not be a good idea! It's just a suggestion).

test/extended/openstack/servergroup.go
pierreprinetti: Question: what would you think about checking the `install-config`'s setting instead of the OpenStack server group setting? I know it's supposed to be the same, but checking the install-config seems more endtoendish to me. WDYT?

test/extended/openstack/servergroup.go
```diff
@@ -79,7 +80,23 @@ var _ = g.Describe("[sig-installer][Feature:openstack] The OpenStack platform",
...
+					o.Expect(host_ids).To(o.HaveLen(len(masterInstanceUUIDs)))
```
rlobillo, line 94: Could you please add an error message on this check? The idea is to check that there are as many elements on host_ids as master nodes, right? I think a descriptive error message will help.

test/extended/openstack/servergroup.go
itzikb-redhat: > as far as I'm aware the "strict" anti-affinity policy is really called `anti-affinity` in OpenStack

yes - you are right :-)

itzikb-redhat: > Question: what would you think about checking the `install-config`'s setting instead of the OpenStack server group setting? I know it's supposed to be the same, but checking the install-config seems more endtoendish to me. WDYT?

I think it's better

itzikb-redhat:
> The Go logic looks good to me.
> 
> Do you mind adding this check to a separate test? That way, we cleanly separate concerns and test results are easier to read.
> 
> I am thinking about something like (just a suggestion!):
> 
> ```go
> g.It("creates Control plane nodes on separate hosts when serverGroupPolicy is anti-affinity", func() {
>     if serverGroupPolicy != "anti-affinity" {
>         e2eskipper.Skipf("This test only applies when serverGroupPolicy is set to anti-affinity")
>     }
>     // [...]
> ```
> 
> If you want to avoid code duplication, consider adding "uuid gathering" to [a `g.BeforeEach` stanza](https://onsi.github.io/ginkgo/#extracting-common-setup-beforeeach) (might not be a good idea! It's just a suggestion).

test/extended/openstack/servergroup.go
itzikb-redhat: @rlobillo Thanks , I'll add

test/extended/openstack/servergroup.go
```diff
@@ -113,3 +147,27 @@ func serverGroupIDsFromName(client *gophercloud.ServiceClient, name string) ([]s
...
+func installConfigFromCluster(client clientcorev1.ConfigMapsGetter) (installConfig, error) {
+	const installConfigName = "cluster-config-v1"
+
+	cm, err := client.ConfigMaps("kube-system").Get(context.Background(), installConfigName, metav1.GetOptions{})
+	if err != nil {
+		return installConfig{}, err
+	}
+
+	var ic installConfig
+	yaml.Unmarshal([]byte(cm.Data["install-config"]), &ic)
```
pierreprinetti, line 170: add error checking here

test/extended/openstack/servergroup.go
```diff
@@ -6,80 +6,114 @@ import (
...
-		dc, err := dynamic.NewForConfig(cfg)
-		o.Expect(err).NotTo(o.HaveOccurred())
+	oc = exutil.NewCLI("openstack")
 
-		g.By("getting the IDs of the Control plane instances and their Server group")
-		masterInstanceUUIDs := make([]interface{}, 0, 3)
-		var serverGroupName string
-		{
+	// OCP 4.5: https://issues.redhat.com/browse/OSASINFRA-1300
+	g.Context("when installed", func() {
```
pierreprinetti, line 36: what is the purpose of this context?

test/extended/openstack/servergroup.go
```diff
@@ -6,80 +6,114 @@ import (
...
 	defer g.GinkgoRecover()
 
-	// OCP 4.5: https://issues.redhat.com/browse/OSASINFRA-1300
-	g.It("creates Control plane nodes in a server group", func() {
-		ctx := context.TODO()
+	var ctx context.Context
+	var computeClient *gophercloud.ServiceClient
+	var dc dynamic.Interface
+	var oc *exutil.CLI
+	var masterInstanceUUIDs []interface{}
```
pierreprinetti, line 30: Please move these declarations to the individual "before" and "it" blocks; it's easier to read when you only declare here the variables you want to share between steps: in this example, `masterInstanceUUIDs` and `serverGroupName`

test/extended/openstack/servergroup.go
```diff
@@ -6,80 +6,114 @@ import (
...
-		o.Expect(err).NotTo(o.HaveOccurred())
-		dc, err := dynamic.NewForConfig(cfg)
-		o.Expect(err).NotTo(o.HaveOccurred())
+	oc = exutil.NewCLI("openstack")
 
-		g.By("getting the IDs of the Control plane instances and their Server group")
-		masterInstanceUUIDs := make([]interface{}, 0, 3)
-		var serverGroupName string
-		{
+	// OCP 4.5: https://issues.redhat.com/browse/OSASINFRA-1300
```
pierreprinetti, line 35: This comment refers to the test "creates Control plane nodes in a server group"

test/extended/openstack/servergroup.go
```diff
@@ -6,80 +6,114 @@ import (
...
+				serverGroup, err := servergroups.Get(computeClient, serverGroupsWithThatName[0]).Extract()
+				o.Expect(serverGroup.Members, err).To(o.ContainElements(masterInstanceUUIDs...))
 			}
-		}
+		})
 
-		g.By("checking the actual members of the Server group")
-		{
-			computeClient, err := client(serviceCompute)
+		g.It("creates Control plane nodes on separate hosts when serverGroupPolicy is anti-affinity", func() {
```
pierreprinetti, line 100: Please add a comment stating the minimum version and the Jira or BZ that implemented/fixed it. In this case, that should be respectively 4.10 and https://issues.redhat.com/browse/OSASINFRA-2507

test/extended/openstack/servergroup.go
```diff
@@ -6,80 +6,114 @@ import (
...
-			ms := dc.Resource(schema.GroupVersionResource{
-				Group:    "machine.openshift.io",
-				Resource: "machines",
-				Version:  "v1beta1",
-			})
-			for _, item := range nodeList.Items {
-				uuid := strings.TrimPrefix(item.Spec.ProviderID, "openstack:///")
-				masterInstanceUUIDs = append(masterInstanceUUIDs, uuid)
+			g.By("getting the IDs of the Control plane instances and their Server group")
+			masterInstanceUUIDs := make([]interface{}, 0, 3)
```
pierreprinetti, line 47: By using `:=`, you declare a new variable. I think you want `=` to use what's declared in the parent block

test/extended/openstack/servergroup.go
```diff
@@ -6,80 +6,114 @@ import (
...
+					instanceServerGroupNameField := getFromUnstructured(res, "spec", "providerSpec", "value", "serverGroupName")
+					o.Expect(instanceServerGroupNameField).NotTo(o.BeNil(), "the Server group name should be present in the Machine definition")
+
+					instanceServerGroupName := instanceServerGroupNameField.(string)
+					o.Expect(instanceServerGroupName).NotTo(o.BeEmpty(), "the Server group name should not be the empty string")
+					if serverGroupName == "" {
+						serverGroupName = instanceServerGroupName
+					} else {
+						o.Expect(instanceServerGroupName).To(o.Equal(serverGroupName), "two Control plane Machines have different serverGroupName set")
+					}
```
pierreprinetti, line 81: This block is really part of the test below, and should not be executed in a before" stanza

test/extended/openstack/servergroup.go
itzikb-redhat: Will renove

test/extended/openstack/servergroup.go
itzikb-redhat: But then I won't be able to use it in the other test

test/extended/openstack/servergroup.go
itzikb-redhat: I'll move

test/extended/openstack/servergroup.go
itzikb-redhat: Done

test/extended/openstack/servergroup.go
itzikb-redhat: Done

test/extended/openstack/servergroup.go
itzikb-redhat: Done

test/extended/openstack/servergroup.go
pierreprinetti: It's fine to redeclare these in each test:

 var ctx context.Context
	var computeClient *gophercloud.ServiceClient
	var dc dynamic.Interface
	var oc *exutil.CLI

test/extended/openstack/servergroup.go
pierreprinetti:
```Go
var ic installConfig
err = yaml.Unmarshal([]byte(cm.Data["install-config"]), &ic)
return ic, err
```
https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshal

itzikb-redhat: Done

test/extended/openstack/servergroup.go
itzikb-redhat: Do you mean to declare them again ? They are used also in BeforeEach

pierreprinetti: Oh. Right, it makes sense to keep here what’s used in the beforeEach 👍

test/extended/openstack/servergroup.go
```diff
@@ -113,3 +155,28 @@ func serverGroupIDsFromName(client *gophercloud.ServiceClient, name string) ([]s
...
+	cm, err := client.ConfigMaps("kube-system").Get(context.Background(), installConfigName, metav1.GetOptions{})
+	if err != nil {
+		return installConfig{}, err
+	}
+
+	var ic installConfig
+	err = yaml.Unmarshal([]byte(cm.Data["install-config"]), &ic)
+	return ic, err
+
+	return ic, nil
```
pierreprinetti, line 181: This line is not reachable; you are already returning nil on line 179 if Unmarshal was successful.

pierreprinetti: You can safely delete this line (and line 180 just above).

pierreprinetti:
Looks amazing. I am setting a /hold while the rehearsal runs. Please remove it (with `/hold cancel`) once `ci/prow/test` is successful.

Thank you!

/approve
/lgtm
/hold

pierreprinetti:
Actually, the rehearsal is required for this to merge. My hold is redundant. I am removing it directly, so that the PR will directly merged if the test is OK.

/hold cancel

PR #12, opened 2021-12-20
Add IP comparison on machineset test

rlobillo:
Covers automation for https://bugzilla.redhat.com/show_bug.cgi?id=2022627

Due to the work done to fix the previous BZ, the Machine Status section now includes all the addresses defined in the OSP instances.

This change adds this check for workers created by a machineset in the existing test "[sig-installer][Feature:openstack] The OpenStack platform on instance creation should follow machineset specs [Suite:openshift/conformance/parallel]".

### DIFF
--- a/test/extended/openstack/servers.go
+++ b/test/extended/openstack/servers.go
@@ -3,9 +3,13 @@ package openstack
 import (
 	"context"
 	"fmt"
+	"net"
+	"reflect"
 	"strconv"
+	"strings"
 	"time"
 
+	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/images"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
@@ -17,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
 	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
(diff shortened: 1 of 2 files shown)


test/extended/openstack/servers.go
```diff
@@ -76,10 +79,14 @@ var _ = g.Describe("[sig-installer][Feature:openstack] The OpenStack platform",
...
 					o.Expect(err).NotTo(o.HaveOccurred())
 					instanceSgs := parseInstanceSgs(instance.SecurityGroups)
+					addressesMachine := getAddressesFromMachine(machine)
+					instanceAddresses, err := parseInstanceAddresses(instance.Addresses)
+					o.Expect(err).NotTo(o.HaveOccurred())
 
 					o.Expect(instanceFlavor.Name).To(o.Equal(flavorMachineSet), "Flavor not matching for instance %q", instance.Name)
 					o.Expect(instanceImage.Name).To(o.Equal(imageMachineSet), "Image not matching for instance %q", instance.Name)
 					o.Expect(instanceSgs).To(o.Equal(sgMachineSets), "SGs not matching for %q", instance.Name)
+					o.Expect(instanceAddresses).To(o.Equal(addressesMachine), "Addresses not matching for instance %q", instance.Name)
```
pierreprinetti, line 89: Can you please mention the BZ in a comment right above this new check?

pierreprinetti: Also, depending on your personal taste, the four lines you added can be replaced with this one-liner:

```Go
o.Expect(parseInstanceAddresses(instance.Addresses)).To(o.Equal(getAddressesFromMachine(machine))
```

the advantage would be that we have all dependent code packed together. Fewer preparatory steps above means that while reading, if I am only interested in `sgMachineSets` for example, I don't have to read about `instanceAddresses`.

but it's really up to you.

pierreprinetti:
I am mildly concerned by this proliferation of checks in a single test. Do you think we should continue checking all MachineSet and Machine properties in a single test, or shall we split each check in its own test?

test/extended/openstack/servers.go
```diff
@@ -172,13 +179,43 @@ func parseInstanceSgs(securityGroups []map[string]interface{}) []string {
...
+}
+
+// parse instance.Addresses object and return a slice of found IPs
+func parseInstanceAddresses(addresses map[string]interface{}) ([]string, error) {
+	result := make([]string, 0, len(addresses)*2)
+	s := fmt.Sprintf("%v", reflect.ValueOf(addresses))
+
+	for _, entry := range strings.Split(s, " ") {
+		if strings.HasPrefix(entry, "addr") {
+			addr := net.ParseIP(strings.Split(entry, ":")[1])
```
pierreprinetti, line 208: In case there is no `:`, the slice resulting from `strings.Split(entry, ":")` will have length `1` and indexing `[1]` will panic.

Also, in order to account for IPv6, we should probably use [`strings.SplitN`](https://pkg.go.dev/strings#SplitN):

```Go
	if strings.HasPrefix(entry, "addr") {
		addrLine := strings.SplitN(entry, ":", 2)
		if len(addrLine) > 1 {
			addr := net.ParseIP(addrLine[1])
			if addr == nil {
				return nil, fmt.Errorf("invalid ip format: %v", addrLine[1])
			}
			result = append(result, addr.String())
		}
	}
```

Alternatively, we can leverage [`strings.TrimPrefix`](https://pkg.go.dev/strings#TrimPrefix). `TrimPrefix` returns the string unchanged if the prefix was not found:
```Go
	if addrValue := strings.TrimPrefix(entry, "addr:"); addrValue != entry {
		addr := net.ParseIP(addrValue)
		if addr == nil {
			return nil, fmt.Errorf("invalid ip format: %v", addrValue)
		}
		result = append(result, addr.String())
	}
	return result, nil
```

rlobillo:
> I am mildly concerned by this proliferation of checks in a single test. Do you think we should continue checking all MachineSet and Machine properties in a single test, or shall we split each check in its own test?

Agree, I moved it to a different testcase, so we are checking the addresses for all the machines, master, workers, belonging to a machine set or not.

test/extended/openstack/servers.go
rlobillo: Thanks!! Done

pierreprinetti:
/approve
/lgtm

Great work, thanks!

PR #17, opened 2022-02-11
[Fix]Put in order the obtained IP lists

rlobillo:
On D/S CI, it is observed that the IP comparison is failing due to a
wrong order in the obtained lists:

Addresses not matching for instance "ostest-mdbrl-worker-0-d9p6c"
Expected
    <[]string | len:2, cap:2>: ["172.17.5.233", "10.196.3.227"]
to equal
    <[]string | len:2, cap:4>: ["10.196.3.227", "172.17.5.233"]

This fix is changing the check from Equal to ConsistOf so the order
is not considered.

### DIFF
--- a/test/extended/openstack/servers.go
+++ b/test/extended/openstack/servers.go
@@ -109,7 +109,7 @@ var _ = g.Describe("[sig-installer][Feature:openstack] The OpenStack platform", 
 				instance, err := servers.Get(computeClient, machine.Get("metadata.annotations.openstack-resourceId").String()).Extract()
 				o.Expect(err).NotTo(o.HaveOccurred(), "Error gathering Openstack info for machine %v", machine.Get("metadata.name"))
 				g.By(fmt.Sprintf("Compare addresses with openstack interfaces for machine %q", instance.Name))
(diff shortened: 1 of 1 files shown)


mdbooth (changes requested):
Couple of things here. Firstly it's not immediately obvious to me why these wouldn't be in the same order. Going forward I would probably want to assert that they are in the same order, although this wasn't true for CAPO prior to 4.10. Do you have an example of the failure?

Secondly, if we *do* decide to ignore order here it would be better to use [Gomega's ConsistOf](https://onsi.github.io/gomega/#consistofelement-interface) here. i.e. something like:

```golang
o.Expect(parseInstanceAddresses(...)).To(o.ConsistOf(getAddressesFromMachine(...)), ...)
```

As well as being simpler, this will also give you a nicer error message if it fails, btw.

mdbooth:
Ok, I can see a potential non-deterministic ordering in MAPO due to iteration over a map, which Go deliberately perturbs. Is this failure non-deterministic by any chance?

For now lets assert any order with ConsistsOf().

mdbooth (approved):
/lgtm

PR #22, opened 2022-05-08
Add a Check for machine status

itzikb-redhat:
Before checking other attributes


rlobillo:
As discussed with @itzikb-redhat, the testcase fails because the machine does not have the SG defined for the worker that is not on 'Running' state. This check will help to identify the root cause of the failure in the test.

/lgtm

PR #27, opened 2022-05-26
do not look for ServiceAccount token secrets

sanchezl:
token secrets go away in kube 1.24


mandre (commented):
/lgtm
/approve

This is commit https://github.com/openshift/origin/commit/985d82356c284318ec133a7260e202ebe7f855a2 from origin.

PR #28, opened 2022-06-19
Use origin as a library

pierreprinetti:
This commit removes most of the code that was vendored from
openshift/origin. Origin is instead referenced as a dependency in
`go.mod`.

Two origin packages remain vendored and must be updated manually:

* the `main` package under `cmd/openshift-tests`;
* the code needed to `go generate` the tests list under `test/extended/util/annotate`.

Both are discussed in README.md.

Implements [OSASINFRA-2902](https://issues.redhat.com//browse/OSASINFRA-2902)


mandre (commented):
/lgtm
/hold for CI

PR #30, opened 2022-06-22
Test consistency on cloud-provider-config

rlobillo:
test_coverage for https://bugzilla.redhat.com/show_bug.cgi?id=2065597

### DIFF
--- a/test/extended/openstack/cloud-provider-config.go
+++ b/test/extended/openstack/cloud-provider-config.go
@@ -1,0 +1,126 @@
+package openstack
+
+import (
+	"context"
+	"strings"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	ini "gopkg.in/ini.v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", func() {
+	defer g.GinkgoRecover()
+
+	var clientSet *kubernetes.Clientset
+	var err error
+
+	g.Context("on cloud provider configuration", func() {
+
+		g.BeforeEach(func() {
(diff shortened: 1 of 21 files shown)


mandre (commented):
Thanks @rlobillo.
I see we're checking the different config maps exist and contain expected entries.
Should we also check that these config maps are updated to the new expected value if we modify the user managed cloud.conf?

For example, changing a param in the `[BlockStorage]` section should update the `openshift-config-managed/kube-cloud-config` and `openshift-cluster-csi-drivers/cloud-conf` config maps but not `openshift-cloud-controller-manager/cloud-conf`.

mandre:
> > Thanks @rlobillo. I see we're checking the different config maps exist and contain expected entries. Should we also check that these config maps are updated to the new expected value if we modify the user managed cloud.conf?
> > For example, changing a param in the `[BlockStorage]` section should update the `openshift-config-managed/kube-cloud-config` and `openshift-cluster-csi-drivers/cloud-conf` config maps but not `openshift-cloud-controller-manager/cloud-conf`.
> 
> AFAIK, changing a param will imply a reconfiguration on the cluster. For example, in the case of changes in the '[LoadBalancer]' section, it will imply a sequential reboot of all the nodes. We are doing that on D/S CI and it is taking ~30 minutes. ([Reference](http://git.app.eng.bos.redhat.com/git/openshift-ir-plugin.git/tree/roles/tests/tasks/change_cloud_cm.yml))

Fair enough, the sync of the config maps should be quick, but I forgot this will trigger the kubelet to restart one by one leading to cluster instability. Forget about my ask then.

test/extended/openstack/cloud-provider-config.go
```diff
@@ -0,0 +1,126 @@
...
+		})
+
+		// https://bugzilla.redhat.com/show_bug.cgi?id=2065597
+		g.It("should haul the user config to the expected config maps", func() {
+
+			type config_params struct {
+				namespace string
+				name      string
+				key       string
+				skip      []string //Slice of properties that are not expected to be present.
```
MaysaMacedo, line 38:
```suggestion
				skip      []string //Slice of properties that are not expected to be present in the internal configMaps
```

rlobillo: Thanks, done.

PR #32, opened 2022-06-23
Bump origin

pierreprinetti:
Update go.mod to https://github.com/openshift/origin/commit/a946e2b9cd8a


mandre (commented):
I could see this quickly becoming cumbersome. We may want to automate the sync with origin on regular basis (weekly?), similar to what mergebot does.

/lgtm
/approve

PR #33, opened 2022-06-23
Disable invariance checks

pierreprinetti:
Do no execute invariant verification while running platform-specific
tests.

This change is expected to reduce noise, while invariant tests are run
alongside the main openshift/conformance run.


pierreprinetti:
I'm happy to reintroduce these checks later on. Until then, I want to make sure they're not failing CI jobs without a clear reason
/approve

PR #34, opened 2022-06-27
Create suite openshift/openstack

pierreprinetti:
Create a suite that only runs OpenStack tests and does not run any
invariant verification.

This is a partial cherry-pick of bdc3567b (https://github.com/openshift/openstack-test/pull/33).


pierreprinetti:
/hold
I'll rather change the test labels (feature->suite) for consistency with the main branch.

PR #43, opened 2022-07-22
Skip use-octavia on cloud provider config test

dulek:
We modify use-octavia option depending on if Kuryr is enabled or not.
This makes the test fail when Kuryr is running. This commit fixes it by
making sure we're not checking that option for matching with user
config.

### DIFF
--- a/test/extended/openstack/cloud-provider-config.go
+++ b/test/extended/openstack/cloud-provider-config.go
@@ -46,23 +46,23 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
 			}
 
 			internalConfigMaps := []config_params{
-				config_params{
+				{
 					namespace: "openshift-config-managed",
 					name:      "kube-cloud-config",
 					key:       "cloud.conf",
-					skip:      nil,
+					skip:      []string{"use-octavia"},
 				},
-				config_params{
+				{
 					namespace: "openshift-cloud-controller-manager",
 					name:      "cloud-conf",
 					key:       "cloud.conf",
-					skip:      []string{"secret-name", "secret-namespace"},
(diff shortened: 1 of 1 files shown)


test/extended/openstack/cloud-provider-config.go
```diff
@@ -46,23 +46,23 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
...
-					skip:      []string{"secret-name", "secret-namespace"},
+					skip:      []string{"secret-name", "secret-namespace", "use-octavia"},
 				},
-				config_params{
+				{
 					namespace: "openshift-cluster-csi-drivers",
 					name:      "cloud-conf",
 					key:       "cloud.conf",
-					skip:      []string{"secret-name", "secret-namespace"},
+					skip:      []string{"secret-name", "secret-namespace", "use-octavia"},
```
MaysaMacedo, line 65: is use-octavia present on the cloud-conf of openshift-cluster-csi-drivers?

mandre: Yes, it does at the moment. The [cloud.conf transformation in cinder-csi-driver-operator](https://github.com/openshift/openstack-cinder-csi-driver-operator/blob/bf20cdb79fcc9c267eab4604ab9291d413ceff1f/pkg/controllers/config/configsync.go#L108) keep existing sections, in this case the `[LoadBalancer]` one.

mandre (commented):
/lgtm

dulek:
Nice, new job is already running! Let's wait for the results.

rlobillo:
It is worth mentioning that this change is necessary to align with the fix implemented for Kuryr at https://bugzilla.redhat.com/show_bug.cgi?id=2074471

We will create a separate test case to specifically check the behavior of the "use-octavia" property.

PR #44, opened 2022-07-25
Check use-octavia property on CCM configMap

rlobillo:
Covers https://bugzilla.redhat.com/show_bug.cgi?id=2074471

### DIFF
--- a/test/extended/openstack/cloud-provider-config.go
+++ b/test/extended/openstack/cloud-provider-config.go
@@ -2,10 +2,12 @@ package openstack
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	ini "gopkg.in/ini.v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -18,6 +20,12 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
 
 	var clientSet *kubernetes.Clientset
 	var err error
+	type configParams struct {
+		namespace string
+		name      string
+		key       string
+		skip      []string //Slice of properties that are not expected to be present in the internal configMaps
+	}
 
 	g.Context("on cloud provider configuration", func() {
 
@@ -31,21 +39,14 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
 		// https://bugzilla.redhat.com/show_bug.cgi?id=2065597
 		g.It("should haul the user config to the expected config maps", func() {
 
-			type config_params struct {
-				namespace string
-				name      string
-				key       string
-				skip      []string //Slice of properties that are not expected to be present in the internal configMaps
-			}
-
-			userConfigMap := config_params{
+			userConfigMap := configParams{
 				namespace: "openshift-config",
 				name:      "cloud-provider-config",
 				key:       "config",
 				skip:      nil,
 			}
 
-			internalConfigMaps := []config_params{
+			internalConfigMaps := []configParams{
 				{
 					namespace: "openshift-config-managed",
 					name:      "kube-cloud-config",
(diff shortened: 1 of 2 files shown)


test/extended/openstack/cloud-provider-config.go
```diff
@@ -104,6 +104,70 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
...
+
+			ccm_cfg, err := getConfig(clientSet,
+				ccmConfigMap.namespace,
+				ccmConfigMap.name,
+				ccmConfigMap.key)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			user_property_value, err := getPropertyValue(section_name, property_name, user_cfg)
+			o.Expect(err).NotTo(o.HaveOccurred(), "Error getting the property %q in section %q on configMap "+
+				"%q (namespace: %q)", property_name, section_name, userConfigMap.name, userConfigMap.namespace)
```
dulek, line 149: We could probably put this into the function itself, so that we don't repeat it in line 152?

test/extended/openstack/cloud-provider-config.go
```diff
@@ -104,6 +104,70 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
...
+			}
+
+			ccmConfigMap := config_params{
+				namespace: "openshift-cloud-controller-manager",
+				name:      "cloud-conf",
+				key:       "cloud.conf",
+				skip:      nil,
+			}
+
+			section_name := "LoadBalancer"
```
dulek, line 125: I know it's a pattern in this file, nevertheless we're starting to mix snake_case with camelCase in the project, this is awful, let's just use camelCase everywhere.

test/extended/openstack/cloud-provider-config.go
```diff
@@ -104,6 +104,70 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
...
+					property_name, section_name, ccmConfigMap.name, ccmConfigMap.namespace)
+				e2e.Logf("  - Property %q with correct value %q on section %q in configMap %q (namespace: %q) with Kuryr NetworkType.",
+					property_name, ccm_property_value, section_name, ccmConfigMap.name, ccmConfigMap.namespace)
+			default:
+				o.Expect(ccm_property_value).Should(o.Equal(user_property_value),
+					"Unexpected value for property %q on section %q. configMap '%q', namespace '%q'.",
+					property_name, section_name, ccmConfigMap.name, ccmConfigMap.namespace)
+				e2e.Logf("  - Property %q with correct value %q on section %q in configMap %q (namespace: %q).",
+					property_name, ccm_property_value, section_name, ccmConfigMap.name, ccmConfigMap.namespace)
+			}
```
dulek, line 169: Why not make it simple?

```
			e2e.Logf("Checking config for networkType detected: %q", network.Status.NetworkType)
			expected := userPropertyValue
			if network.Status.NetworkType == "Kuryr" {
				expected = "true"
			}

			o.Expect(ccmPropertyValue).Should(o.Equal(expected),
				"Unexpected value for property %q on section %q. configMap '%q', namespace '%q' with NetworkType=%s.",
				propertyName, sectionName, ccmConfigMap.name, ccmConfigMap.namespace, network.Status.NetworkType)
			e2e.Logf("  - Property %q with correct value %q on section %q in configMap %q (namespace: %q) with NetworkType=%s.",
				propertyName, ccmPropertyValue, sectionName, ccmConfigMap.name, ccmConfigMap.namespace)
```

test/extended/openstack/cloud-provider-config.go
```diff
@@ -124,3 +188,16 @@ func getConfig(kubeClient kubernetes.Interface, namespace string, cm_name string
 	}
 	return cfg, nil
 }
+
+// return an harmonized string with the value of a property defined inside a section.
+func getPropertyValue(section_name string, property_name string, cfg *ini.File) (string, error) {
```
dulek, line 193: I think I'd add error handling here directly so that we don't need to repeat it when running this function. It's private anyway.

test/extended/openstack/cloud-provider-config.go
```diff
@@ -104,6 +104,70 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
...
+				"%q (namespace: %q)", property_name, section_name, userConfigMap.name, userConfigMap.namespace)
+
+			ccm_property_value, err := getPropertyValue(section_name, property_name, ccm_cfg)
+			o.Expect(err).NotTo(o.HaveOccurred(), "Error getting the property %q in section %q on configMap "+
+				"%q (namespace: %q)", property_name, section_name, ccmConfigMap.name, ccmConfigMap.namespace)
+
+			e2e.Logf("Checking config for networkType detected: %q", network.Status.NetworkType)
+			switch network.Status.NetworkType {
+			case "Kuryr":
+				o.Expect(ccm_property_value).Should(o.Equal("true"),
```
dulek, line 158: Wait, I have an impression it should be "false" when Kuryr is running. Yet this passes tests, so I'm confused now.

dulek (changes requested):
Looks good, some style remarks and one doubt inline.

test/extended/openstack/cloud-provider-config.go
dulek: Ah, camelCase is also preferred by golang in general: https://go.dev/doc/effective_go#mixed-caps

test/extended/openstack/cloud-provider-config.go
dulek: Yeah, seems like test does nothing.

test/extended/openstack/cloud-provider-config.go
rlobillo: I prefer not to through errors on functions and handle them on the test itself. I believe this is better for debugging failing testcases.

test/extended/openstack/cloud-provider-config.go
rlobillo: I changed it all along the file. Thanks! Will do like that from now on

test/extended/openstack/cloud-provider-config.go
rlobillo: Thanks for the suggestion. I changed the logic a little bit. I included also the "enabled" property in the checks.

test/extended/openstack/cloud-provider-config.go
rlobillo: It is private but can be reused for other future tests in the file. As I mention before, I prefer to see the exceptions on the test itself, it is better for debugging.

test/extended/openstack/cloud-provider-config.go
rlobillo: According to the logs I am writting and the different tests I did, it is working fine:

```
$ make openstack-tests && KUBECONFIG=~/ostest/auth/kubeconfig ./openstack-tests run-test "[sig-installer][Suite:openshift/openstack] The Openshift on cloud provider configuration should update use-octavia and enabled properties depending on the NetworkType [Suite:openshift/conformance/parallel]"
go build -o openstack-tests ./cmd/openshift-tests
[BeforeEach] [Top Level]
  /home/stack/openstack-test/vendor/github.com/openshift/origin/test/extended/util/test.go:61
[BeforeEach] on cloud provider configuration
  /home/stack/openstack-test/test/extended/openstack/cloud-provider-config.go:31
STEP: preparing openshift dynamic client
[It] should update use-octavia and enabled properties depending on the NetworkType [Suite:openshift/conformance/parallel]
  /home/stack/openstack-test/test/extended/openstack/cloud-provider-config.go:109
Jul 26 16:15:54.032: INFO: Checking config for networkType detected: "Kuryr"
Jul 26 16:15:54.032: INFO: Properties with correct values on section "LoadBalancer" in configMap "cloud-conf" (namespace: "openshift-cloud-controller-manager") with NetworkType=Kuryr.
Jul 26 16:15:54.032: INFO: - Properties (["use-octavia" "enabled"]): Values (["true" "false"])

```

rlobillo: the ccm config map is setting use-octavia=True and enabled=False. This is what I understood from the BZ and the PR fixing it (https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/183)

test/extended/openstack/cloud-provider-config.go
dulek: Okay.

test/extended/openstack/cloud-provider-config.go
```diff
@@ -66,55 +67,125 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
...
+				o.Expect(err).NotTo(o.HaveOccurred(), "Error getting the property %q in section %q on configMap "+
+					"%q (namespace: %q)", propertyName, sectionName, ccmConfigMap.name, ccmConfigMap.namespace)
+				ccmPropertyValues = append(ccmPropertyValues, value)
+			}
+
+			g.By(fmt.Sprintf("Checking properties for networkType detected: %q", network.Status.NetworkType))
+			var expectedValues []string
+			if network.Status.NetworkType == "Kuryr" {
+				//Fixed values regardless the content of userCfg.
+				expectedValues = []string{"true", "false"}
```
dulek, line 160: Okay, so IIUC this only makes sense once we make a switch to use CCM. If so I would wait with merging this test until that happens as we're cautious with the possibility of it slipping again.

rlobillo: In my local tests, I am not testing with external CCM enabled in the cluster, however the CCM configMap is configured. If you want to use this test to confirm that the fix implemented for https://bugzilla.redhat.com/show_bug.cgi?id=2074471 remains in place, I would merge it. But, if you think that behaviour could change, I am ok waiting until 4.12 enables CCM by default.

dulek: Okay, I see the point, it's more about CCM bug even though CCM is just tech preview. Cool, I'm fine with this then, but I'll let @MaysaMacedo take a look too.

dulek:
/lgtm

Leaving /approve for a second pair of eyes.

test/extended/openstack/cloud-provider-config.go
```diff
@@ -66,55 +67,125 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The Openshift", f
...
+			} else {
+				//Should map the values configured on the userCfg.
+				for _, propertyName := range propertyNames {
```
MaysaMacedo, line 163: When Kuryr is not used, we fix the value of use-octavia as well, to always be true since lbaasv2 is not supported. Perhaps this test can be updated to take that into account?
https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/openstack/openstack.go#L85-L93

rlobillo: Hello @MaysaMacedo, in such a case the property use-octavia is useless on the cloud-provider-config?

What should happen if use-octavia = false or the property is not present in cloud-provider-config?

MaysaMacedo: cluster-cloud-controller-manager (CCCMO) enforces that configuration.
I'm proposing the removal of it from the installer as is not needed https://github.com/openshift/installer/pull/5834

rlobillo: Great! Thanks, I will update the test accordingly. No need to check the cloud-provider-config then.

on the eCCM cm, for Kuryr: 

`expectedValues = []string{"true", "false"}`

for any other networkType:

`expectedValues = []string{"true", "#UNDEFINED"}`

MaysaMacedo: I think we still need it. If the user changes the user managed configmap, the internal one will continue to enforce the appropriate values.

rlobillo: Thanks @MaysaMacedo I updated the name of the test and updated the expected values for NetworkType != Kuryr to fixed values. 

I also tested locally with OpenShiftSDN and Kuryr and in both cases the testcase is passing.

Please take a look and let me know if you think something is missing.

MaysaMacedo:
/lgtm
While I saw the test passed, I didn't see any logs of the steps within the test. Is this expected?

rlobillo:
> /lgtm While I saw the test passed, I didn't see any logs of the steps within the test. Is this expected?

Hello @MaysaMacedo 

Checking [0], the steps are not written for any test. I checked for already running tests and it is not there either. If you want to see the steps you need to run the test isolated, like:

`openshift-test run-test "<name_of_the_test>"`

I think this is the expected behaviour.

[0] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_openstack-test/44/pull-ci-openshift-openstack-test-main-test-kuryr/1556886463394615296/artifacts/test-kuryr/openstack-test-openstack/artifacts/e2e.log

MaysaMacedo:
okay, thanks!
/approve

MaysaMacedo:
/hold
There is something wrong with a kuryr test case, which is unrelated to this test scenario. I'm taking a look

PR #45, opened 2022-08-04
Add a test to check for machineset leftovers ports

itzikb-redhat:
Create a machineset with a bogus serverGroupID
Delete the machineset
Check that there is no leftover port associated with the deleted machineset


rlobillo:
Hello @itzikb-redhat I think you can move the test of servers.go as I think it is suitable there. For the file I created, I try to give "openstack" perspective (servers, servergroups, volumes...) and the operation you are testing is related to the creating of a new server. What do you think?

itzikb-redhat:
> Hello @itzikb-redhat I think you can move the test to **servers.go** file. For the files I created until now, I tried to give "openstack" perspective (servers, servergroups, volumes...) and the operation you are testing is related to the creation of a new server. What do you think?

I think it's more related to machinesets but we can discuss

rlobillo (commented):
Great job @itzikb-redhat 

I think it would be very useful in this complex testcase to make use of **g.By**. With this utility it is easier for maintainers to understand the different stages in the testcase. Would you mind to include them in the test? This will help for review too.

Also, in the **Expect** commands, it is possible to include a failure message: 

`o.Expect(err).NotTo(o.HaveOccurred(), "Error unmarshaling new MachineSet Specs")
`

I think that is very useful to get a clear view of the failure by only checking the logs.

[0] https://onsi.github.io/ginkgo/#documenting-complex-specs-by

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,135 @@
...
+
+		})
+
+		g.It("should not have leftovers ports", func() {
+
+			skipUnlessMachineAPIOperator(dc, clientSet.CoreV1().Namespaces())
+			g.By("fetching worker machineSets")
+			var networkClient *gophercloud.ServiceClient
+			var rawBytes []byte
+			var t machineproviderv1.OpenstackProviderSpec
```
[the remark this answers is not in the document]
rlobillo: I was referring to line 47 only. :)

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,135 @@
...
+	g.Context("after deletion of a machineset", func() {
+		g.BeforeEach(func() {
+			g.By("preparing openshift dynamic client")
+			cfg, err := e2e.LoadConfig()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			dc, err = dynamic.NewForConfig(cfg)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			clientSet, err = e2e.LoadClientset()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			g.By("preparing openstack client")
```
MaysaMacedo, line 46: this line and the next one are unnecessary

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,135 @@
...
+
+			networkClient, err := client(serviceNetwork)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			machineSets, err := listWorkerMachineSets(dc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			if len(machineSets) == 0 {
+				e2eskipper.Skipf("Expects at least one worker machineset. Found none.")
+			}
+			xconfig, err := config.GetConfig()
```
MaysaMacedo, line 68: Is it needed to get the config again if it's already available at line 40 or it's a different one?

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,135 @@
...
+
+			err = machinev1.AddToScheme(scheme.Scheme)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			err = configv1.AddToScheme(scheme.Scheme)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			newMachinesetParams := framework.BuildMachineSetParams(rclient, 1)
+			rawBytes, err = json.Marshal(newMachinesetParams.ProviderSpec.Value)
+
+			err = json.Unmarshal(rawBytes, &t)
```
MaysaMacedo, line 82: what t represents? a machineset?

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,135 @@
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			t.ServerGroupName = ""
+			t.ServerGroupID = "boo"
+			newMachinesetParams.Name = "bogus1-worker"
+			s, err := json.Marshal(t)
+			newMachinesetParams.ProviderSpec.Value.Raw = s
+
+			o.Expect(err).NotTo(o.HaveOccurred())
```
MaysaMacedo, line 91: it might be better to move this line to after 88 so we can easily identify which err is being checked

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,135 @@
...
+			ms, err := framework.CreateMachineSet(rclient, newMachinesetParams)
```
MaysaMacedo, line 92:
```suggestion
                        g.By("Create machineSet with new name and server group")
			ms, err := framework.CreateMachineSet(rclient, newMachinesetParams)
```

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,135 @@
...
+			// Check if an orphaned port related to the delete machineset exists
+			allPages, err := ports.List(networkClient, portListOpts).AllPages()
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			allPorts, err := ports.ExtractPorts(allPages)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			portID := ""
+			for _, port := range allPorts {
+				if strings.Contains(port.Name, newMachinesetParams.Name) {
```
MaysaMacedo, line 128: is the name of the port exactly a fragment of the machineset name or it should be equal?
If it's equal, we could filter the ports query by that name as well

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,135 @@
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+			ms, err := framework.CreateMachineSet(rclient, newMachinesetParams)
+
+			clusterID := ms.Spec.Template.ObjectMeta.Labels["machine.openshift.io/cluster-api-cluster"]
+
+			_, err = framework.GetMachineSet(rclient, ms.Name)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			//Delete the machineset
+			e2e.Logf("Waiting for 10 seconds before deleting machineset %v", ms.Name)
```
MaysaMacedo, line 100: What happens if 10 seconds was not enough to create the machine? Perhaps we should poll instead

MaysaMacedo (commented):
Is this test an outcome of a bugzilla?

itzikb-redhat:
> 

@MaysaMacedo [BZ2073398](https://bugzilla.redhat.com/show_bug.cgi?id=2073398)

test/extended/openstack/machinesets.go
itzikb-redhat: Done

test/extended/openstack/machinesets.go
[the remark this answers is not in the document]
itzikb-redhat: Done

MaysaMacedo:
> > 
> 
> @MaysaMacedo [BZ2073398](https://bugzilla.redhat.com/show_bug.cgi?id=2073398)

@itzikb-redhat It would be good to add a comment to the code, linking that bz. Thanks

test/extended/openstack/machinesets.go
itzikb-redhat: I'm changing to e2e.LoadConfig , anyway cfg is not accessible here

test/extended/openstack/machinesets.go
itzikb-redhat: I modified the name to newProviderSpec, It's the providerspec of the machineset

test/extended/openstack/machinesets.go
itzikb-redhat: Done

test/extended/openstack/machinesets.go
itzikb-redhat: Add a similar one

test/extended/openstack/machinesets.go
itzikb-redhat: The ports contains the string but it's not exact

itzikb-redhat:
> > > 
> > 
> > 
> > @MaysaMacedo [BZ2073398](https://bugzilla.redhat.com/show_bug.cgi?id=2073398)
> 
> @itzikb-redhat It would be good to add a comment to the code, linking that bz. Thanks

Done

pierreprinetti:
Rebased on `main`

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,145 @@
...
+			rclient, err = runtimeclient.New(cfg, runtimeclient.Options{})
+			o.Expect(err).NotTo(o.HaveOccurred(), "Error creating a runtime client")
+
+			err = machinev1.AddToScheme(scheme.Scheme)
+			o.Expect(err).NotTo(o.HaveOccurred(), "Failed to add Machine to scheme")
+			err = configv1.AddToScheme(scheme.Scheme)
+			o.Expect(err).NotTo(o.HaveOccurred(), "Failed to add Config to scheme")
+
+			newMachinesetParams := framework.BuildMachineSetParams(rclient, 1)
+			rawBytes, err = json.Marshal(newMachinesetParams.ProviderSpec.Value)
```
MaysaMacedo, line 78: this err also needs to be handled

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,145 @@
...
+			for retries < maxRetries {
+				_, err = framework.GetMachineSet(rclient, ms.Name)
+				if err != nil {
+					retries += 1
+					time.Sleep(time.Second * delay)
+				} else {
+					break
+				}
+			}
+			o.Expect(err).NotTo(o.HaveOccurred(), "Failed to get the new Machineset")
```
MaysaMacedo, line 109: if an error happened once and eventually the machine was found, will we still have an err here and fail when it shouldn't?

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,145 @@
...
+
+			g.By("Deleting the new machineset")
+			framework.DeleteMachineSets(rclient, ms)
+			_, err = framework.GetMachineSet(rclient, ms.Name)
```
MaysaMacedo, line 113: we probably need to poll for machine not found in this case

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,145 @@
...
+			newProviderSpec.ServerGroupID = "boo"
+			newMachinesetParams.Name = "bogus1-worker"
+			newProviderSpecJson, err := json.Marshal(newProviderSpec)
+			o.Expect(err).NotTo(o.HaveOccurred(), "Failed to marshal new Machineset provider spec")
+			newMachinesetParams.ProviderSpec.Value.Raw = newProviderSpecJson
+
+			g.By("Create a new machineSet with bogus server group ID")
+			ms, err := framework.CreateMachineSet(rclient, newMachinesetParams)
+			o.Expect(err).NotTo(o.HaveOccurred(), "Failed to create a Machineset")
+			defer DeleteMachinesetsDefer(rclient, ms)
```
MaysaMacedo, line 93: is there a need for a defer if the machine is explicitly deleted in the test?

test/extended/openstack/machinesets.go
itzikb-redhat: If at the last iteration of the loop there was an error then it raises it , even there wasn't then it breaks out of the loop and the expect should be fine

test/extended/openstack/machinesets.go
itzikb-redhat: We need it in case the test errs before the machineset is deleted

test/extended/openstack/machinesets.go
itzikb-redhat: Done

test/extended/openstack/utils.go
```diff
@@ -122,3 +133,30 @@ func difference(a []string, b []string) []string {
...
+				break
+			}
+		} else {
+			if err == nil {
+				retries += 1
+				time.Sleep(time.Second * delay)
+			} else {
+				break
+			}
+		}
```
dulek, line 159: Whoa, what a code duplication,. why? How about:

	if err != nil == shouldExist {
		retries += 1
		time.Sleep(time.Second * delay)
	} else {
		break
	}

Please double-check me if the logic is identical.

You should also check the error type here if it's NotFound.

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,146 @@
...
+			// retries := 1
+			// for retries < maxRetries {
+			// 	_, err = framework.GetMachineSet(rclient, ms.Name)
+			// 	if err != nil {
+			// 		retries += 1
+			// 		time.Sleep(time.Second * delay)
+			// 	} else {
+			// 		break
+			// 	}
+			// }
```
dulek, line 109: This has to go before merge.

test/extended/openstack/machinesets.go
```diff
@@ -0,0 +1,146 @@
...
+			newMachinesetParams := framework.BuildMachineSetParams(rclient, 1)
+			rawBytes, err = json.Marshal(newMachinesetParams.ProviderSpec.Value)
+			o.Expect(err).NotTo(o.HaveOccurred(), "Error marshaling new MachineSet Provider Spec")
+
+			err = json.Unmarshal(rawBytes, &newProviderSpec)
+			o.Expect(err).NotTo(o.HaveOccurred(), "Error unmarshaling new MachineSet Provider Spec")
+
+			newProviderSpec.ServerGroupName = ""
+			newProviderSpec.ServerGroupID = "boo"
+			newMachinesetParams.Name = "bogus1-worker"
```
dulek, line 85: You should have a random name here instead.

dulek (changes requested):
Looks pretty good, minor comments inline.

test/extended/openstack/utils.go
itzikb-redhat: agree! And logic is ok

test/extended/openstack/machinesets.go
itzikb-redhat: Done

test/extended/openstack/machinesets.go
itzikb-redhat: Done

test/extended/openstack/utils.go
itzikb-redhat: About the Notfound - I'm checking it in the test itself, do you think I should add the logic in the function?

dulek: Nah, okay, I see NotFound check now.

itzikb-redhat:
> Looks good. One more thing is that it's a good practice to put vendoring changes into a separate commit in the PR, so that later on we can revert the test without necessarily reverting the `go mod` update. Please update that.

Done. Please let me know if it seems ok

dulek:
@itzikb-redhat: Please do not retest this, looks like we've found an interference between two tests:

```
{  fail [github.com/openshift/openstack-test/test/extended/openstack/servers.go:114]: Error gathering Openstack info for machine bogus-2881xjhrl-mbq69
Unexpected error:
    <gophercloud.ErrDefault404>: {
        ErrUnexpectedResponseCode: {
            BaseError: {DefaultErrString: "", Info: ""},
            URL: "https://compute.rdo.mtl2.vexxhost.net/v2.1/0693e2bb538c42b79a49fe6d2e61b0fc/servers/",
            Method: "GET",
            Expected: [200, 203],
            Actual: 404,
            Body: "{\"message\": \"The resource could not be found.<br /><br />\\n\\n\\n\", \"code\": \"404 Not Found\", \"title\": \"Not Found\"}",
        },
    }
    Resource not found
occurred

I guess [here](https://github.com/openshift/openstack-test/blob/61409a88ec92dddabb836be062cbd2ed77a248ab/test/extended/openstack/servers.go#L104-L118) either a new Machine should be created, or only installer-created worker nodes should be used. Or we should just catch 404 and ignore it, later on checking if we've found at least one Machine. Please note that filtering to only have Machines in "Provisioned" status will not work, the state of the machine can easily change between fetching list and calling OpenStack.

itzikb-redhat:
@dulek Thanks, @rlobillo saw that issue. We'll propose a patch.

MaysaMacedo:
/hold
CCM was not deployed, likely depends on https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/202

MaysaMacedo:
/hold cancel
dependent PR was merged

MaysaMacedo:
/approve remove
seems other test is being impacted

itzikb-redhat:
So "[sig-installer][Suite:openshift/openstack] The OpenStack platform on instance creation should include the addresses on the machine specs [Suite:openshift/conformance/parallel]"
 failed

The updated PS should solve it for now but I wonder if the test shouldn't be modified to reflect the bug , i.e. to add a FIP to a machine and then check the addresses of the machine instead of going over all the machines

PR #54, opened 2022-10-05
OCPBUGS-2023: Skip image check when booting from volume

mandre:
When the instance was booted from a volume, it does not have the image field populated. This caused the `on instance creation should follow machineset specs` test to fail.
Skip the image check when it is empty.


pierreprinetti:
/approve
Thanks!

PR #57, opened 2022-11-04
UDP-test: Fix lbMethod comparison

rlobillo:
Util function 'getPropertyValue' provides the result in lower case, so when lbMethod is set in the configMap, the validation with the pool value fails.

This change is for ensuring lower case comparison always.


rlobillo:
Already covered on https://github.com/openshift/openstack-test/pull/58

PR #58, opened 2022-11-04
Add udp-lb testcase for ovn provider

rlobillo:
Covers partially the epic https://issues.redhat.com/browse/OSASINFRA-2753

This test creates an UDP lb when OVN provider is configured on the cluster and checks the created Openstack resources. It also sends 100 requests to the LoadBalancer Service and analyze the pods' responses so that the e2e functionality is verified.

This test will be skipped for Kuryr NetworkType and if cloudProviderConfig is configuring a different lb-provider than OVN.

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/gophercloud/gophercloud"
@@ -33,6 +34,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb] The Openstack
 	var loadBalancerClient *gophercloud.ServiceClient
 	var networkClient *gophercloud.ServiceClient
 	var clientSet *kubernetes.Clientset
+	availableLbProvidersUnderTests := [2]string{"Amphora", "OVN"}
 
 	g.BeforeEach(func() {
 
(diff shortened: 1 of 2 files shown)


MaysaMacedo:
do we expect the configuration to be manually updated before running the test?
Or should the test update the configuration to use the OVN driver?

rlobillo:
Hello @MaysaMacedo 

Yes, we expect the configuration to be manually updated before running the test. The configuration change takes around 30-40 minutes and it can affect the run of the rest of the tests. 

In D/S CI we are doing the cloud-provider-config changes and running the openstack-tests sequentially: https://code.engineering.redhat.com/gerrit/gitweb?p=rhos-qe-jenkins.git;hb=refs%2Fchanges%2F16%2F434216%2F4;f=jobs%2Fdefaults%2Fstages%2Fir_openshift_udplb_tests_run.groovy.inc

MaysaMacedo:
I opened a ticket to vexxhost about the issue seen here

test/extended/openstack/loadbalancers.go
```diff
@@ -54,101 +56,103 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb] The Openstack
...
-			podName, err := getPodNameThroughUdpLb(svcIp, svcPort, "hostname")
-			if err != nil {
-				e2e.Logf("Error detected while accessing the LoadBalancer service on try %d: %q", i, err)
-			} else {
-				results[podName]++
+			g.By("Creating Openshift resources")
+			labels := map[string]string{"app": "udp-lb-default-dep"}
+			depName := "udp-lb-default-dep"
+			svcName := "udp-lb-default-svc"
+			testDeployment := e2edeployment.NewDeployment(depName, 2, labels, "udp-test",
```
MaysaMacedo, line 79: it would be good to add a step of g.By("Creating deployment")

test/extended/openstack/loadbalancers.go
```diff
@@ -54,101 +56,103 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb] The Openstack
...
+			}}
+			deployment, err := clientSet.AppsV1().Deployments(oc.Namespace()).Create(context.TODO(),
+				testDeployment, metav1.CreateOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+			err = e2edeployment.WaitForDeploymentComplete(clientSet, deployment)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			jig := e2eservice.NewTestJig(clientSet, oc.Namespace(), svcName)
+			jig.Labels = labels
+			svc, err := jig.CreateLoadBalancerService(5*time.Minute, func(svc *v1.Service) {
```
MaysaMacedo, line 96: it would be good to add a step of g.By("Creating Load-Balancer")

test/extended/openstack/loadbalancers.go
```diff
@@ -54,101 +56,103 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb] The Openstack
...
+				svc.Spec.Ports = []v1.ServicePort{{Protocol: v1.ProtocolUDP, Port: 8082, TargetPort: intstr.FromInt(8081)}}
+				svc.Spec.Selector = labels
+			})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("Checks from openshift perspective")
+			o.Expect(svc.GetAnnotations()["loadbalancer.openstack.org/load-balancer-id"]).ShouldNot(o.BeEmpty(),
+				"load-balancer-id annotation missing")
+			loadBalancerId := svc.GetAnnotations()["loadbalancer.openstack.org/load-balancer-id"]
```
MaysaMacedo, line 105: can this be defined before line 103 and then use the variable on the expect check?

test/extended/openstack/loadbalancers.go
```diff
@@ -54,101 +56,103 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb] The Openstack
...
+			o.Expect(svc.Status.LoadBalancer.Ingress[0].IP).ShouldNot(o.BeEmpty(), "FIP missing on svc Status")
+			o.Expect(svc.Spec.ExternalTrafficPolicy).Should(o.Equal(v1.ServiceExternalTrafficPolicyTypeCluster),
+				"Unexpected ExternalTrafficPolicy on svc specs")
+
+			g.By("Checks from openstack perspective")
+			lb, err := octavialoadbalancers.Get(loadBalancerClient, loadBalancerId).Extract()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			o.Expect(lb.Provider).Should(o.Equal(strings.ToLower(lbProviderUnderTest)), "Unexpected provider in the Openstack LoadBalancer")
+
+			svcIp := svc.Status.LoadBalancer.Ingress[0].IP
```
MaysaMacedo, line 115: ditto before line 106

test/extended/openstack/loadbalancers.go
```diff
@@ -54,101 +56,103 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb] The Openstack
...
-		o.Expect(err).NotTo(o.HaveOccurred())
-		o.Expect(lb.Provider).Should(o.Equal("amphora"), "Unexpected provider in the Openstack LoadBalancer")
-
-		svcIp := svc.Status.LoadBalancer.Ingress[0].IP
-		svcPort := fmt.Sprint(svc.Spec.Ports[0].Port)
-		fip, err := getFipbyFixedIP(networkClient, lb.VipAddress)
-		o.Expect(err).NotTo(o.HaveOccurred())
-		o.Expect(fip.FloatingIP).Should(o.Equal(svcIp), "Unexpected floatingIp in the Openstack LoadBalancer")
-		o.Expect(lb.Pools).Should(o.HaveLen(1), "Unexpected number of pools on Openstack LoadBalancer %q", lb.Name)
+	for _, i := range availableLbProvidersUnderTests {
```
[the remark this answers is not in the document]
rlobillo: I don't know why but your proposal did not work. I tried that, but if I don't set the variable later, the loop is not working as expected. I found this approach on origin repo, for example: https://github.com/openshift/origin/blob/master/test/extended/apiserver/apply.go#L57

test/extended/openstack/loadbalancers.go
```diff
@@ -54,101 +56,103 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb] The Openstack
...
+			g.By("Checking loadBalancer provider")
+			cloudProviderConfig, err := getConfig(oc.AdminKubeClient(),
+				"openshift-config",
+				"cloud-provider-config",
+				"config")
+			o.Expect(err).NotTo(o.HaveOccurred())
+			lbProvider, err := getClusterLoadBalancerSetting("lb-provider", cloudProviderConfig)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if lbProvider != strings.ToLower(lbProviderUnderTest) {
+				e2eskipper.Skipf("Test not applicable for LoadBalancer provider different than %s: %q", lbProviderUnderTest, lbProvider)
```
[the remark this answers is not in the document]
rlobillo: Sure! will do

test/extended/openstack/loadbalancers.go
rlobillo: I need to know if the key is present before assigning it, right?

test/extended/openstack/loadbalancers.go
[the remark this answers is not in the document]
dulek: Instead of looping like this shouldn't we just check the LB provider that's configured and then when LB is created in Octavia checking if it has correct provider?

test/extended/openstack/loadbalancers.go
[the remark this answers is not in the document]
rlobillo: Done

test/extended/openstack/loadbalancers.go
rlobillo: Done

test/extended/openstack/loadbalancers.go
rlobillo: Done

test/extended/openstack/loadbalancers.go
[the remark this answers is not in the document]
dulek: They're testing different providers really. But I can't argue with the concept of a separate test if you need it like this for your tools. The loop and skip just feels a bit weird here.

Do you need both tests to appear in every result set, even if skipped? If so, I guess `for` loop is the way to go. Otherwise you could check LB provider before `g.It(fmt.Sprintf("should create an UDP %s LoadBalancer when an UDP svc with type:LoadBalancer is created on Openshift", lbProviderUnderTest)`.

test/extended/openstack/loadbalancers.go
dulek: You'll get empty string if it's not assigned, so it's safe to just save it into the var and then do the checks: https://go.dev/play/p/hP_VtvL_K8t

test/extended/openstack/loadbalancers.go
dulek: Well, the string will be empty in worst case so no problem there, but you should check for length of the Ingress array, otherwise if it's empty, you'll end up with some golang panic here.

test/extended/openstack/loadbalancers.go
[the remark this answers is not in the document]
rlobillo: Thanks! Done

test/extended/openstack/loadbalancers.go
rlobillo: Thanks! Done

MaysaMacedo:
/hold cancel
I will leave the approve for someone else. Maybe @dulek?

PR #59, opened 2022-11-15
Add udp shared lb testcases

rlobillo:
Covers partially the epic https://issues.redhat.com/browse/OSASINFRA-2753

This test creates an UDP lb svc and created a second one re-using the same openstack loadbalancer. It checks that the openstack loadbalancer is configured appropiatedly and the e2e connectivity works.

This test will be skipped for Kuryr NetworkType and if max-shared-lb has a value lower than 2.

PR #61, opened 2022-11-16
Add pre-created fip testcase

rlobillo:
Covers partially the epic https://issues.redhat.com/browse/OSASINFRA-2753

Test that creates a FIP on openstack and uses it for creating an UDP svc type:LoadBalancer that sets on its spec the pre-created FIP.

This test will be skipped for Kuryr NetworkType and will run with both amphora and OVN provider.

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -12,7 +12,10 @@ import (
 	octavialoadbalancers "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
 	octaviamonitors "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/external"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
 	exutil "github.com/openshift/origin/test/extended/util"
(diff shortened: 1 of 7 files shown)


test/extended/openstack/loadbalancers.go
```diff
@@ -141,6 +144,59 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb] The Openstack
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+		g.By(fmt.Sprintf("FIP created: %s", fip.FloatingIP))
+		defer floatingips.Delete(networkClient, fip.ID)
+
+		g.By(fmt.Sprintf("Create Openshift resources using the FIP %s on network %s", fip.FloatingIP, fip.FloatingNetworkID))
+		labels := map[string]string{"app": "udp-lb-precreatedfip-dep"}
+		depName := "udp-lb-precreatedfip-dep"
+		svcName := "udp-lb-precreatedfip-svc"
+		svcPort := int32(8022)
+		testDeployment := e2edeployment.NewDeployment(depName, 2, labels, "udp-test",
```
MaysaMacedo, line 164: Seems this deployment creation will be happening in every new test suit,  might be good to extract it to a function

test/extended/openstack/loadbalancers.go
```diff
@@ -233,3 +289,29 @@ func isLbMethodApplied(lbMethod string, results map[string]int, pods *v1.PodList
...
+		return nil, err
+	}
+
+	allNetworks, err := networks.ExtractNetworks(allPages)
+	if err != nil {
+		return nil, err
+	}
+
+	networkIds := make([]string, len(allNetworks))
+	for x, network := range allNetworks {
```
MaysaMacedo, line 312: We need to jump any ipv6 only networks as FIPs are not supported with ipv6

dulek (commented):
Looks good please take a look at Maysa's remarks.

test/extended/openstack/loadbalancers.go
rlobillo: I'm addressing this on https://github.com/openshift/openstack-test/pull/59 and will adapt the rest of PRs accordingly as soon as it gets merged

rlobillo:
@MaysaMacedo @dulek last UDP LB test to merge in master for this round. Please take a look.

/override ci/prow/test-kuryr

test/extended/openstack/loadbalancers.go
```diff
@@ -419,3 +478,56 @@ func createTestDeployment(depName string, labels map[string]string, replicas int
...
+				return "", err
+			}
+			for _, networkSubnet := range network.Subnets {
+				subnet := getSubnet(networkSubnet, subnetList)
+				if subnet != nil {
+					if subnet.IPVersion == 4 {
+						return network.ID, nil
+					}
+				} else {
+					return network.ID, nil
```
dulek, line 517: You still wants to return if subnet is not accessible?

test/extended/openstack/loadbalancers.go
```diff
@@ -419,3 +478,56 @@ func createTestDeployment(depName string, labels map[string]string, replicas int
 	}}
 	return testDeployment
 }
+
+// GetFloatingNetworkID returns a floating network ID.
+func GetFloatingNetworkID(client *gophercloud.ServiceClient) (string, error) {
```
dulek, line 483: So this mimics how the cloud provider gets the ID of the floating IP network? What happens if it's forced to be something else in the configuration? IMO test should check the config too. Or you should pass your own network as an annotation when creating the service, but then it's probably not guaranteed the connectivity will always work if you choose something inappropriate for the setup.

dulek (changes requested):

test/extended/openstack/loadbalancers.go
rlobillo: Yes, in fact that is the typical case: The external network is created by the openstack admin with `--no-share true` so the tenant cannot inspect the external subnets. Please refer to https://github.com/kubernetes/cloud-provider-openstack/pull/2014

test/extended/openstack/loadbalancers.go
rlobillo: Exactly, these functions are highly inspired on https://github.com/openshift/cloud-provider-openstack/blob/master/pkg/util/openstack/network.go#L53

The intention of this test is to validate that user can set a previously created FIP on the svc definition, so CCM does not need to create it at all, and it just makes use of it.

Therefore, what is defined on the cloud-provider-config should be irrelevant in this case.

test/extended/openstack/loadbalancers.go
dulek: Alright!

test/extended/openstack/loadbalancers.go
dulek: @rlobillo: What if there's only one "correct" FIP subnet that can be accessed from where the tests are run and it is set in the config? If autodetection takes incorrect one the connectivity check will fail. I'm just trying to find the corner cases so that when some job with a specific configuration will be run we won't need to debug why this test fails.

rlobillo: @dulek: OK! the test will use the subnet  on cloud-provider-config in case it's configured and, if not, auto-detect. I'm on it.

dulek:
/hold cancel

Looking good!

PR #72, opened 2023-01-03
Parameterize the Load Balancer service creation timeout

rlobillo:
Due to [1], the LBs are created secuentially in OpenStack
so if the LB tests in openstack-test run in parallel it
would be needed to increase the loadBalancerServiceTimeout
to a higher value.

With the serialization of the LB tests [2] there is no need
to increase loadBalancerServiceTimeout but this commit leaves
the parameter in case it needs to be changed in the future.

[1] https://issues.redhat.com/browse/OCPBUGS-6069
[2] https://github.com/openshift/openstack-test/pull/78


mandre:
/retitle Parameterize the Load Balancer service creation timeout

mandre (commented):
/lgtm
/approve

PR #74, opened 2023-01-17
Fix comparison of machine addresses with openstack instance addresses

mdbooth:
Firstly this test was broken because skipUnlessMachineAPIOperator is broken and was executing against a UPI cluster. We address this by only fetching OpenStack instances for machines which exist. If no machines exist (because it's UPI) the test trivially succeeds.

Secondly we improve the error reporting. We update the gomega matchers to operate on the actual data. This means that if they don't match the non-matching data will be included in the error message.

### DIFF
--- a/hack/verify.sh
+++ b/hack/verify.sh
@@ -27,7 +27,14 @@ verify_generated() {
 	declare new='test/extended/util/annotate/generated/zz_generated.annotations.go'
 	cp "$new" "$old"
 	go generate ./test/extended
-	diff "$old" "$new"
+	if ! diff "$old" "$new"; then
+            echo 'Generated files are out of date'
+            echo 'Run "make update"'
+            rm $old
+            return 1
+        fi
+        rm $old
+        return 0
 }
 
 declare run=0 failed=0 junit_testcases=''
--- a/test/extended/openstack/servers.go
+++ b/test/extended/openstack/servers.go
@@ -104,26 +104,40 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 		})
 
 		// https://bugzilla.redhat.com/show_bug.cgi?id=2022627
-		g.It("should include the addresses on the machine specs", func() {
+		g.It("should report all openstack instance addresses on the corresponding Machine object", func() {
+			// N.B. A UPI installation will not have any Machine objects. This test correctly trivially succeeds in that case because it
+			// iterates over Machine objects which exist.
 
-			skipUnlessMachineAPIOperator(dc, clientSet.CoreV1().Namespaces())
 			g.By("fetching machines")
 			machines, err := getMachines(dc)
 			o.Expect(err).NotTo(o.HaveOccurred())
 
-			foundMachines := 0
+			// Create 2 maps of machine names to addresses: one from the machine specs, and one from OpenStack
+			machineIPs := make(map[string][]string)
+			openstackIPs := make(map[string][]string)
+
+			// Fetch IP addresses of OpenStack instances corresponding to the machines
 			for _, machine := range machines {
-				g.By(fmt.Sprintf("Gather Openstack attributes for machine %q", machine.Get("metadata.name")))
+				machineName := machine.Get("metadata.name").String()
+				machineIPs[machineName] = getAddressesFromMachine(machine)
+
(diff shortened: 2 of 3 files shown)


test/extended/openstack/servers.go
```diff
@@ -302,3 +315,10 @@ func skipUnlessMachineAPIOperator(dc dynamic.Interface, c coreclient.NamespaceIn
 	})
 	o.Expect(err).NotTo(o.HaveOccurred())
 }
+
+func mapKeys[K comparable, V any](m map[K]V) (keys []K) {
```
pierreprinetti, line 319: first generics in openstack-test!

pierreprinetti:
Your solution for this test looks fine, however it seems a bit too implicit WRT the UPI case in particular.
The crux of it all being:
> Incidentally `skipUnlessMachineAPIOperator` seems broken

Do you see a way to address this more broadly? Maybe by making `skipUnlessMachineAPIOperator` skip if there are no ControlPlane machines in the cluster? Or maybe, UPI should have a manifest deactivating MAO in CVO and we check that here?

mdbooth:
> Your solution for this test looks fine, however it seems a bit too implicit WRT the UPI case in particular. The crux of it all being:
> 
> > Incidentally `skipUnlessMachineAPIOperator` seems broken
> 
> Do you see a way to address this more broadly? Maybe by making `skipUnlessMachineAPIOperator` skip if there are no ControlPlane machines in the cluster? Or maybe, UPI should have a manifest deactivating MAO in CVO and we check that here?

Honestly I've always preferred code which handles empty lists by trivially iterating over them rather than adding a special case. This test asserts that all openshift machines must have corresponding openstack instances with the same addresses. This is true for both UPI and IPI. That's extremely tidy to me.

I haven't looked at other uses of skipUnlessMachineAPIOperator yet. We will need to fix it properly if we have any tests that genuinely need to special case it, but perhaps we can even delete it? Less special cases is more better. Lets treat them one by one.

pierreprinetti:
> is test asserts that all openshift machines must have corresponding openstack instances with the same addresses.

That's what the test does, not what it says ("The OpenStack platform - on instance creation - should include the addresses on the machine specs")

mdbooth:
> > is test asserts that all openshift machines must have corresponding openstack instances with the same addresses.
> 
> That's what the test does, not what it says ("The OpenStack platform - on instance creation - should include the addresses on the machine specs")

Good point. I fixed the test description to describe what it's actually doing. This hasn't changed, btw.

I also added a comment pointing out that it works for UPI too.

pierreprinetti:
Thank you.
/lgtm

mdbooth:
Verify is failing, but I don't know what the output means:

```
> ./hack/verify.sh


Jan 17 18:50:19.872: INFO: Enabling in-tree volume drivers
758c758
<       "[sig-installer][Suite:openshift/openstack] The OpenStack platform on instance creation should include the addresses on the machine specs": "",
---
>       "[sig-installer][Suite:openshift/openstack] The OpenStack platform on instance creation should report all openstack instance addresses on the corresponding Machine object": "",
```

Is it complaining that the test name changed?

mdbooth:
> Verify is failing, but I don't know what the output means:
> 
> ```
> > ./hack/verify.sh
> 
> 
> Jan 17 18:50:19.872: INFO: Enabling in-tree volume drivers
> 758c758
> <       "[sig-installer][Suite:openshift/openstack] The OpenStack platform on instance creation should include the addresses on the machine specs": "",
> ---
> >       "[sig-installer][Suite:openshift/openstack] The OpenStack platform on instance creation should report all openstack instance addresses on the corresponding Machine object": "",
> ```
> 
> Is it complaining that the test name changed?

Found it. It's complaining that the generated file is out of date. I'm going to add a better error message along the lines of the gofmt one.

hack/verify.sh
```diff
@@ -27,7 +27,14 @@ verify_generated() {
 	declare new='test/extended/util/annotate/generated/zz_generated.annotations.go'
 	cp "$new" "$old"
 	go generate ./test/extended
-	diff "$old" "$new"
+	if ! diff "$old" "$new"; then
+            echo 'Generated files are out of date'
+            echo 'Run "Go generate ./..."'
```
pierreprinetti, line 32:
```
echo 'Run "make update"'

pierreprinetti (approved):

PR #76, opened 2023-01-19
Fix maxSharedLb test skip condition expression

eurijon:
The test should be skipped when maxSharedLb is less than 2, as Octavia won't allow sharing a load balancer and the test case would fail.


test/extended/openstack/loadbalancers.go
```diff
@@ -165,7 +165,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb] The Openstack
 			o.Expect(err).NotTo(o.HaveOccurred())
 			maxSharedLb, err := strconv.Atoi(setting)
 			o.Expect(err).NotTo(o.HaveOccurred())
-			if 2 < maxSharedLb {
+			if 2 > maxSharedLb {
```
mandre, line 168: Nit, `if maxSharedLb < 2` is easier to read and better matches that the error on the next line reports.

mandre (commented):
/lgtm
/approve

PR #78, opened 2023-01-27
OCPBUGS-6712: Serialise LB tests

pierreprinetti:
Due to https://issues.redhat.com/browse/OCPBUGS-6069 the LB type svc
creation time depends on the number of LBs being created in parallel,
thus on the number of LB tests running in parallel. By making them run
serially the svc creation time won't depend on the number of tests
running in parallel.

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -34,7 +34,7 @@ import (
 	imageutils "k8s.io/kubernetes/test/utils/image"
 )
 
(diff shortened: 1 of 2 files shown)


pierreprinetti:
It looks like they actually ran serially: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_openstack-test/78/pull-ci-openshift-openstack-test-main-test/1618928447885676544/artifacts/test/openstack-test-openstack/build-log.txt

According to the logs, the next test seems to only start after the previous has completed.

PR #82, opened 2023-02-06
Bug OCPBUGS-7105: Fix MachineSet test

MaysaMacedo:
When using BYON with Proxy the name of the nodes network might not contain the ID of cluster, consequently the test is unable to identify the network. This commit fixes the issue by gathering the information from the primarySubnet in the MachineSet field.


mandre (commented):
@MaysaMacedo feel free to remove the hold if you're satisfied with the CI results.
/lgtm
/approve

PR #77, opened 2023-01-20
[release-4.11] Rebase on Origin

pierreprinetti:
Rebase on origin's release-4.11 commit https://github.com/openshift/origin/commit/f61477ff79e35ffb263bc047bb70dc4e3c4611d2

PR #83, opened 2023-02-10
OSASINFRA-3088: Restore gophercloud/utils to the previous version

mandre:
It was downgraded by mistake in https://github.com/openshift/openstack-test/pull/77


pierreprinetti:
/approve
/lgtm

Thank you; I must have downgraded it by mistake indeed.

PR #84, opened 2023-02-10
OCPBUGS-7310: Do not consider deleted instance as error

mandre:
The `OpenStack platform on instance creation should report all openstack instance addresses on the corresponding Machine object` test list all machines in the openshift cluster then fetches information from OpenStack for each one of them. Because the test doesn't run in isolation, it's affected by other tests that run at the same time and it makes it flaky.

We had a check that looked for errors getting info for a machine that did not take into account deleted machines. Move this check at the right place.


pierreprinetti:
If we get false negatives, we might consider serialising the test instead. However, I agree that there is a cost to it.

/approve
/lgtm

PR #88, opened 2023-02-16
OCPBUGS-7310: Fix test checking if Machine addresses match VMs

dulek:
The aforementioned test blindly tries to check all the Machines in the cluster, assuming each have a VM in OpenStack. That might not be true as in other tests we create Machines that are supposed to fail provisioning and never produce a VM. This commit fixes this by adding several precautions:

* We only consider machines having addresses set in .status.
* We skip machines without openstack-resourceId annotation.
* We skip checking machines that do not have VM in OpenStack (could have been deleted between we fetched the Machines and did OpenStack query).

### DIFF
--- a/test/extended/openstack/servers.go
+++ b/test/extended/openstack/servers.go
@@ -119,10 +119,15 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 			// Fetch IP addresses of OpenStack instances corresponding to the machines
 			for _, machine := range machines {
 				machineName := machine.Get("metadata.name").String()
-				machineIPs[machineName] = getAddressesFromMachine(machine)
+				machineAddresses := getAddressesFromMachine(machine)
+				machineResourceID := machine.Get("metadata.annotations.openstack-resourceId").String()
+				if len(machineAddresses) == 0 || len(machineResourceID) == 0 {
+					// Only consider machines that have addresses and a VM in OpenStack.
+					continue
+				}
 
 				g.By(fmt.Sprintf("Gather Openstack attributes for machine %q", machineName))
-				instance, err := servers.Get(computeClient, machine.Get("metadata.annotations.openstack-resourceId").String()).Extract()
+				instance, err := servers.Get(computeClient, machineResourceID).Extract()
 
 				var gerr gophercloud.ErrDefault404
 				if !errors.As(err, &gerr) {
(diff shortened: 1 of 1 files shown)


rlobillo:
Hi @dulek,

I think we should not skip machines without openstack-resourceId annotation. That is one of the main checks in the test. 


Isnt "only consider machines having addresses set in .status" enough for stabilizing the test?

dulek:
> Hi @dulek,
> 
> I think we should not skip machines without openstack-resourceId annotation. That is one of the main checks in the test.
> 
> Isnt "only consider machines having addresses set in .status" enough for stabilizing the test?

But if `openstackResourceId == ""`, then the try to fetch the VM from Nova will imminently fail.

rlobillo:
> > Hi @dulek,
> > I think we should not skip machines without openstack-resourceId annotation. That is one of the main checks in the test.
> > Isnt "only consider machines having addresses set in .status" enough for stabilizing the test?
> 
> But if `openstackResourceId == ""`, then the try to fetch the VM from Nova will imminently fail.

Please ignore my comment. The openstack-resourceId annotation is checked on the previous test https://github.com/openshift/openstack-test/blob/85a5fb21736881854d30ef3dfd96daef760806de/test/extended/openstack/servers.go#L88

Sorry for the noise

test/extended/openstack/servers.go
```diff
@@ -119,7 +119,12 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
...
-				machineIPs[machineName] = getAddressesFromMachine(machine)
+				machineAddresses := getAddressesFromMachine(machine)
+				machineResourceID := machine.Get("metadata.annotations.openstack-resourceId").String()
+				if len(machineAddresses) == 0 || len(machineResourceID) == 0 {
+					// Only consider machines that have addresses and a VM in OpenStack.
+					continue
+				}
 
 				g.By(fmt.Sprintf("Gather Openstack attributes for machine %q", machineName))
 				instance, err := servers.Get(computeClient, machine.Get("metadata.annotations.openstack-resourceId").String()).Extract()
```
mandre, line 130:
```suggestion
				instance, err := servers.Get(computeClient, machineResourceID).Extract()
```

mandre (commented):
Just a small nit, otherwise LGTM.

test/extended/openstack/servers.go
dulek: Done!

mandre (commented):
/lgtm
/approve

PR #89, opened 2023-02-17
Add UDP LB loadBalancerSourceRanges service test

eurijon:
The test:

- Creates a UDP LB type svc with loadBalancerSourceRanges set to '9.9.9.9/32', meaning only the traffic coming from such source IP is allowed. This IP address has been chosen as there is a low chance it belongs to the client where the test runs from.
- Checks the loadBalancerSourceRanges is reflected in the Openstack load balancer listener (in allowed_cidrs).
- Checks the requests to the service do not succeed
- Updates the service removing the loadBalancerSourceRanges spec
- Checks the loadBalancerSourceRanges change is reflected in the Openstack load balancer listener (allowed_cidrs: 0.0.0.0/0 means no restriction)
- Checks the requests to the service succeed now

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/gophercloud/gophercloud"
+	octavialisteners "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
 	octavialoadbalancers "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
 	octaviamonitors "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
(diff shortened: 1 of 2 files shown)


test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,121 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		listeners, err := octavialisteners.ExtractListeners(allPages)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		// Check there is only one listener in the LB
+		o.Expect(listeners).Should(o.HaveLen(1), "More than one listener found for LB '%s': %v", loadBalancerId, listeners)
+		listenerId := listeners[0].ID
+		e2e.Logf("LB listener ID: %s", listenerId)
+
+		// Check allowed_cidrs in the LB listener matches with the svc spec
+		slicesEqual := stringSlicesEqual(listeners[0].AllowedCIDRs, []string{drop_sourcerange})
```
rlobillo, line 439: I think you can use omega.ConsistOf here instead of the stringSlicesEqual function: https://onsi.github.io/gomega/#working-with-collections

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,121 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		// Check there is no connectivity
+		g.By("Connectivity checks - there should be no connectivity")
+		var connNumber int = 5
+		g.By(fmt.Sprintf("accessing the service %d times from outside and storing the name of the pods answering", connNumber))
+		results := make(map[string]int)
+		for i := 0; i < connNumber; i++ {
+			// https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/README.md#netexec
+			podName, err := getPodNameThroughUdpLb(svcIp, fmt.Sprintf("%d", svcPort), "hostname")
+			if err != nil {
+				e2e.Logf("Error detected while accessing the LoadBalancer service on try %d: %q", i, err)
```
rlobillo, line 452: maybe we can re-phrase this as this is not an error but an expected behaviour?

test/extended/openstack/loadbalancers.go
```diff
@@ -630,3 +746,39 @@ func getMembersMatchingStatus(client *gophercloud.ServiceClient, pool *pools.Poo
 	}
 	return members, nil
 }
+
+// Active wait until allowed_cidrs is updated in a given load balancer listener
+func waitUntilLbListenerAllowedCidrsIsUpdated(client *gophercloud.ServiceClient, interval time.Duration, timeout time.Duration, listenerId string, expectedCidrs []string) error {
```
rlobillo, line 751: This is new for me but I found something that can make this poll stuff: https://onsi.github.io/gomega/#making-asynchronous-assertions. Not sure if that can be applicable here, maybe something like:

`o.Eventually(60, getAllowedCIDRs(client, listernerId)).Should(o.Consistof(AllowedCIDRs))`

test/extended/openstack/loadbalancers.go
eurijon: Done

test/extended/openstack/loadbalancers.go
eurijon: Nice, I changed it

test/extended/openstack/loadbalancers.go
eurijon: Nice catch! I replaced it and it works very well, I think Gomega Eventually and Consistently are really useful here :)

rlobillo:
/lgtm

@dulek @MaysaMacedo PTAL

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+	// Only valid with the Octavia amphora provider. The OVN-Octavia provider does not
+	// support allowed_cidrs in the LB (https://issues.redhat.com/browse/OCPBUGS-2789)
+	g.It("should limit service access on an UDP Amphora LoadBalancer when an UDP LoadBalancer svc setting the loadBalancerSourceRanges spec is created on Openshift", func() {
+
+		// Skipped for OVN-Octavia provider
+		skipIfNotLbProvider("Amphora", cloudProviderConfig)
+
+		g.By("Creating Openshift deployment")
+		depName := "udp-lb-sourceranges-dep"
+		svcName := "udp-lb-sourceranges-svc"
```
dulek, line 396: We only use `svcName` once, I'd probably put it directly in the `NewTestJig()` call.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		svcPort := int32(8082)
+		replicas := int32(2)
```
dulek, line 398: Same here.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		labels := map[string]string{"app": depName}
+		drop_sourcerange := "9.9.9.9/32"
```
dulek, line 400: Why "drop"? It seems to me it's more like "allow".

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+			svc.Spec.Selector = labels
+			svc.Spec.LoadBalancerSourceRanges = []string{drop_sourcerange}
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Checks from openshift perspective")
+		loadBalancerId := svc.GetAnnotations()["loadbalancer.openstack.org/load-balancer-id"]
+		o.Expect(loadBalancerId).ShouldNot(o.BeEmpty(), "load-balancer-id annotation missing")
+		e2e.Logf("LB ID: %s", loadBalancerId)
+		o.Expect(svc.Spec.LoadBalancerSourceRanges).Should(o.Equal([]string{drop_sourcerange}))
```
dulek, line 422: This one feels really redundant, I don't think we should check here if OpenShift API saves objects correctly.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		e2e.Logf("Service IP: %s", svcIp)
+
+		g.By("Checks from openstack perspective")
+		allPages, err := octavialisteners.List(loadBalancerClient, octavialisteners.ListOpts{LoadbalancerID: loadBalancerId}).AllPages()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		listeners, err := octavialisteners.ExtractListeners(allPages)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		// Check there is only one listener in the LB
+		o.Expect(listeners).Should(o.HaveLen(1), "More than one listener found for LB '%s': %v", loadBalancerId, listeners)
```
dulek, line 434: What if it's 0? We should change the log here.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		listenerId := listeners[0].ID
+		e2e.Logf("LB listener ID: %s", listenerId)
+
+		// Check allowed_cidrs in the LB listener matches with the svc spec
+		o.Expect(listeners[0].AllowedCIDRs).Should(o.ConsistOf([]string{drop_sourcerange}), "Unexpected allowed_cidrs in Openstack LoadBalancer Listener '%s'", listenerId)
```
dulek, line 439: We should log the value of `listeners[0].AllowedCIDRs`. Will it log it?

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		e2e.Logf("Expected allowed_cidrs: '%v'", allowAllAllowedCidrs)
+		o.Eventually(func() []string {
+			lbListener, err := octavialisteners.Get(loadBalancerClient, listenerId).Extract()
+			if err != nil {
+				e2e.Logf("Error ocurred: %v, trying next iteration", err)
+				return []string{}
+			}
+			e2e.Logf("Found AllowedCIDRs: %v", lbListener.AllowedCIDRs)
+			return lbListener.AllowedCIDRs
+
```
dulek, line 478: Weird newline.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		e2e.Logf("Expected ProvisioningStatus: '%s'", activeProvisioningStatus)
+		o.Eventually(func() string {
+			lbListener, err := octavialisteners.Get(loadBalancerClient, listenerId).Extract()
+			if err != nil {
+				e2e.Logf("Error ocurred: %v, trying next iteration", err)
+				return ""
+			}
+			e2e.Logf("Found ProvisioningStatus: %s", lbListener.ProvisioningStatus)
+			return lbListener.ProvisioningStatus
+
```
dulek, line 493: Same here.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		}, "10s", "1s").Should(o.Equal(activeProvisioningStatus), "Didn't find the expected ProvisioningStatus '%s'", activeProvisioningStatus)
```
dulek, line 494: Just 10 seconds of waiting? This feels like something that could fail often with Amphora.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		e2e.Logf("Found expected ProvisioningStatus '%s' in Openstack LoadBalancer Listener '%s'", activeProvisioningStatus, listenerId)
+
+		// Check there is still only one listener in the LB and that the Id matches with the previously stored one
+		allPages, err = octavialisteners.List(loadBalancerClient, octavialisteners.ListOpts{LoadbalancerID: loadBalancerId}).AllPages()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		listeners, err = octavialisteners.ExtractListeners(allPages)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(listeners).Should(o.HaveLen(1), "More than one listener found for LB '%s': %v", loadBalancerId, listeners)
```
dulek, line 502: Same here, message is misleading when len=0.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		pods, err := oc.KubeClient().CoreV1().Pods(oc.Namespace()).List(context.Background(), metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		var successConnCount int
+		for _, pod := range pods.Items {
+			e2e.Logf("Checking if pod '%s' has been accessed...", pod.Name)
+			val, ok := results[pod.Name]
+			o.Expect(ok).To(o.BeTrue(), "Pod '%s' not accessed", pod.Name)
+			e2e.Logf("Pod '%s' has been successfully accessed", pod.Name)
+			successConnCount += val
+		}
```
dulek, line 528: Why do we want to make sure all pods were accessed? I don't see value, IMO if we're super unlucky this may make this test flaky.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 			e2e.Logf("Pod %s successfully accessed through svc loadbalancer on ip %s and port %d", podName, fip.FloatingIP, svcPort)
 		})
 	}
+
+	// Only valid with the Octavia amphora provider. The OVN-Octavia provider does not
+	// support allowed_cidrs in the LB (https://issues.redhat.com/browse/OCPBUGS-2789)
+	g.It("should limit service access on an UDP Amphora LoadBalancer when an UDP LoadBalancer svc setting the loadBalancerSourceRanges spec is created on Openshift", func() {
```
dulek, line 389: Why UDP? loadBalancerSourceRanges is independent from protocol, so I'd consider using TCP as it's just working better with Octavia in any version. ;)

dulek (changes requested):

test/extended/openstack/loadbalancers.go
eurijon: Done

test/extended/openstack/loadbalancers.go
eurijon: Done

test/extended/openstack/loadbalancers.go
eurijon: Agree, I've removed it

test/extended/openstack/loadbalancers.go
eurijon: Right, I've changed it to:
```
o.Expect(listeners).Should(o.HaveLen(1), "Unexpected number of listeners for LB '%s'", loadBalancerId)
```
The error log will show the listeners object and the expected length:
i.e (ignore the values as I changed the expected number of listeners in order to trigger the error):

```
fail [/home/stack/openstack-test/test/extended/openstack/loadbalancers.go:452]: Unexpected number of listeners for LB '36305d21-e3dc-4870-84f3-f4cdcc872b13'
Expected
    <[]listeners.Listener | len:1, cap:4>: [
        {
            ID: "a027f57d-4c12-496e-80e7-5d156105d3ab",
            ProjectID: "30d0f65255d94819b724e76a8f81e0ea",
            Name: "listener_0_kube_service_kubernetes_e2e-test-openstack-qczsq_udp-lb-sourceranges-svc",
            Description: "",
            Protocol: "UDP",
            ProtocolPort: 8082,
            DefaultPoolID: "ffe26c35-d849-4238-be41-325045848cf8",
            Loadbalancers: [
                {
                    ID: "36305d21-e3dc-4870-84f3-f4cdcc872b13",
                },
            ],
            ConnLimit: -1,
            SniContainerRefs: [],
            DefaultTlsContainerRef: "",
            AdminStateUp: true,
            Pools: nil,
            L7Policies: [],
            ProvisioningStatus: "ACTIVE",
            TimeoutClientData: 50000,
            TimeoutMemberData: 50000,
            TimeoutMemberConnect: 5000,
            TimeoutTCPInspect: 0,
            InsertHeaders: {},
            AllowedCIDRs: ["9.9.9.9/32"],
        },
    ]
to have length 2
```

test/extended/openstack/loadbalancers.go
eurijon: It's being logged when the gomega asserion fails. i.e (ignore the values again):
```
fail [/home/stack/openstack-test/test/extended/openstack/loadbalancers.go:458]: Unexpected allowed_cidrs in Openstack LoadBalancer Listener '8aa7c328-540b-4215-b79d-1ddbb9716cb7'
Expected
    <[]string | len:1, cap:4>: ["9.9.9.9/32"]
to consist of
    <[]string | len:1, cap:1>: ["10.10.10.10/32"]
the missing elements were
    <[]string | len:1, cap:1>: ["10.10.10.10/32"]
the extra elements were
    <[]string | len:1, cap:1>: ["9.9.9.9/32"]
Ginkgo exit error 1: exit with code 1
```

test/extended/openstack/loadbalancers.go
eurijon: Removed

test/extended/openstack/loadbalancers.go
eurijon: Removed

test/extended/openstack/loadbalancers.go
eurijon: I added an active wait because it's not immediate, but it takes around 1 sec in most cases (with amphora).
```
Mar 10 04:25:10.729: INFO: Expected allowed_cidrs: '[0.0.0.0/0]'
Mar 10 04:25:10.792: INFO: Found AllowedCIDRs: [9.9.9.9/32]
Mar 10 04:25:11.858: INFO: Found AllowedCIDRs: [0.0.0.0/0]
Mar 10 04:25:11.858: INFO: Found expected allowed_cidrs '[0.0.0.0/0]' in Openstack LoadBalancer Listener '75935039-b393-4349-acb5-1fa3adfb86f3'
```

test/extended/openstack/loadbalancers.go
eurijon: Changed to:
```
o.Expect(listeners).Should(o.HaveLen(1), "Unexpected number of listeners for LB '%s'", loadBalancerId)
```

test/extended/openstack/loadbalancers.go
eurijon: I've removed it as we don't want it to be flaky due to the balancing between pods, but I'm keeping the check of number of successful connections.

test/extended/openstack/loadbalancers.go
eurijon: There is no such test for TCP yet so we prefer to test it even if it's with UDP. This service is not of type ETP:local so there should not be any issue as it won't rely on loadbalancer monitors.

dulek:
Looks good with me, leaving /lgtm to @itzikb-redhat as his remark seems unanswered.

test/extended/openstack/loadbalancers.go
```diff
@@ -382,6 +383,152 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+		// Check there is no connectivity
+		g.By("Connectivity checks - there should be no connectivity")
+		var connNumber int = 5
+		g.By(fmt.Sprintf("Trying to access the service %d times from outside and storing the name of the pods answering", connNumber))
+		results := make(map[string]int)
+		for i := 0; i < connNumber; i++ {
+			// https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/README.md#netexec
+			podName, err := getPodNameThroughUdpLb(svcIp, fmt.Sprintf("%d", svcPort), "hostname")
+			if err != nil {
+				e2e.Logf("No connectivity (as expected) while accessing the LoadBalancer service on try %d: %q", i, err)
```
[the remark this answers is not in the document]
eurijon: There rest of UDP tests check the connectivity to a newly created services, so this one checks the loadBalancerSourceRanges policy is enforced in a newly created service (and not only after the svc is updated).

I reused an already defined function to check the connectivity and added a few extra lines to check there is no connectivity. As this is the only test where that's being checked I would implement it if it's needed for more tests.

eurijon:
> Looks good with me, leaving /lgtm to @itzikb-redhat as his remark seems unanswered.

Thanks a lot for your review

eurijon:
I've added some margin (80%) to the number of successful connections check to avoid test flakiness.

test/extended/openstack/loadbalancers.go
dulek: But that's a log of API value check above, not "Found expected ProvisioningStatus". I'm afraid it might take longer for Amphora to actually process this sometimes. I couldn't find this log in the gate.

dulek (changes requested):
Let's discuss the 10 seconds timeout a bit more.

test/extended/openstack/loadbalancers.go
eurijon: My fault... I pasted the incorrect log lines, these are the correct logs:

```
Mar 10 11:44:52.185: INFO: Expected ProvisioningStatus: 'ACTIVE'
Mar 10 11:44:52.344: INFO: Found ProvisioningStatus: PENDING_UPDATE
Mar 10 11:44:53.512: INFO: Found ProvisioningStatus: PENDING_UPDATE
Mar 10 11:44:54.665: INFO: Found ProvisioningStatus: PENDING_UPDATE
Mar 10 11:44:56.105: INFO: Found ProvisioningStatus: ACTIVE
Mar 10 11:44:56.105: INFO: Found expected ProvisioningStatus 'ACTIVE' in Openstack LoadBalancer Listener '539337ad-8fcf-4ba1-9165-ccd495c6e6d9'
```

The log is from the test that failed in the gate: https://prow.ci.openshift.org/61bfa262-82c4-4dc5-86f9-9006a1e609df
and it took less than 4 seconds to update the listener for an Amphora.

Do you think that increasing the timeout, to 20s for instance, is reasonable?

dulek: Yes, I think this strongly depends on how stressed the env is and may take longer than that. Let's maybe put 30 seconds in there just to be safe here from false positives.

eurijon: Done!

PR #75, opened 2023-01-19
Use Origin's release-4.10 as a dependency

pierreprinetti:
Using commit https://github.com/openshift/origin/commit/1d82b3ccec679a0c15413d7b6068f1dac42e26d0 as a base.

PR #91, opened 2023-02-22
OSASINFRA-3108: Bump gophercloud/utils

mandre:
We need a more recent version for self-signed certificates. It was downgraded by mistake in https://github.com/openshift/openstack-test/pull/75.


pierreprinetti:
Thank you!

PR #93, opened 2023-02-22
OSASINFRA-3079: Disable LB tests in proxy envs

mandre:
These tests currently fail in because we use BYON and not Floating IPs.


EmilienM:
awesome! I was looking at how to do it :)
nicely done
/lgtm
/approve

PR #98, opened 2023-03-27
Add EgressIP CloudPrivateIPConfig test for OVN-Kubernetes

rlobillo:
The test checks the port/floating IP assignment in Openstack for the EgressIP is kept after an EgressIP failover.

It runs only for OVN-Kubernetes networkType and when at least 2 workers exist on the given OCP cluster.

### DIFF
--- a/test/extended/openstack/egressip.go
+++ b/test/extended/openstack/egressip.go
@@ -1,0 +1,479 @@
+package openstack
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"os"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	cloudnetwork "github.com/openshift/client-go/cloudnetwork/clientset/versioned"
+	exutil "github.com/openshift/origin/test/extended/util"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+)
+
+const (
+	egressIPConfigAnnotationKey = "cloud.network.openshift.io/egress-ipconfig"
+	egressAssignableLabelKey    = "k8s.ovn.org/egress-assignable"
+)
+
+var _ = g.Describe("[sig-installer][Suite:openshift/openstack][egressip] An egressIP", func() {
+
+	var networkClient *gophercloud.ServiceClient
+	var clientSet *kubernetes.Clientset
+	var err error
+	var infraID string
+	var workerNodeList *corev1.NodeList
+	var cloudNetworkClientset cloudnetwork.Interface
+	oc := exutil.NewCLI("openstack")
+
+	g.BeforeEach(func() {
+
+		g.By("Loading the kubernetes clientset")
+		clientSet, err = e2e.LoadClientset()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		networkClient, err = client(serviceNetwork)
+		o.Expect(err).NotTo(o.HaveOccurred(), "Error creating an openstack network client")
+
+		// TODO revert once https://issues.redhat.com/browse/OSASINFRA-3079 is resolved
+		proxy, err := oc.AdminConfigClient().ConfigV1().Proxies().Get(context.Background(), "cluster", metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if proxy.Status.HTTPProxy != "" {
+			e2eskipper.Skipf("Test not applicable for proxy setup")
+		}
+
+		infrastructure, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		infraID = infrastructure.Status.InfrastructureName
+
+		g.By("Getting the worker node list")
+		workerNodeList, err = clientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
+			LabelSelector: "node-role.kubernetes.io/worker",
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+	})
+
+	g.It("attached to a floating IP should be kept after EgressIP node failover with OVN-Kubernetes NetworkType", func() {
+
+		g.By("Getting the network type")
+		networkType, err := getNetworkType(oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if networkType != NetworkTypeOVNKubernetes {
+			e2eskipper.Skipf("Test not applicable for '%s' NetworkType (only valid for '%s')", networkType, NetworkTypeOVNKubernetes)
+		}
+
+		// Skip based on number of required worker nodes
+		minWorkerNum := 2
+		g.By(fmt.Sprintf("Check the number of worker nodes (should be %d at least)", minWorkerNum))
+		e2e.Logf("Detected number of workers: '%d'", len(workerNodeList.Items))
+		if len(workerNodeList.Items) < minWorkerNum {
+			e2eskipper.Skipf("Skipping test as there are '%d' worker node(s) and minimum required is '%d'", len(workerNodeList.Items), minWorkerNum)
+		}
+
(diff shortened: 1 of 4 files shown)


pierreprinetti:
This PR adds the oc cli to the test image: https://github.com/openshift/release/pull/37830

pierreprinetti:
Yay! It’s still failing! But the oc tool is apparently available now. Way to go.

rlobillo:
yes, that's on me, I can continue now, thank you!

test/extended/openstack/egressip.go
```diff
@@ -0,0 +1,470 @@
...
+		g.By(fmt.Sprintf("Labeling the secondary node '%s' with '%s'", secondaryWorker.Name, egressAssignableLabelKey))
+		e2e.AddOrUpdateLabelOnNode(clientSet, secondaryWorker.Name, egressAssignableLabelKey, "dummy")
+		defer e2e.RemoveLabelOffNode(clientSet, secondaryWorker.Name, egressAssignableLabelKey)
+
+		// Check the assigned node in CloudPrivateIPConfig is kept with the primary EgressIP worker node
+		g.By("Checking the CloudPrivateIPConfig object and the assigned node after adding a second EgressIP assignable node")
+		cpicAssignedNode, err := getCpipAssignedNode(cloudNetworkClientset, egressIPAddrStr)
+		o.Expect(err).NotTo(o.HaveOccurred(), "Could not find the assigned node in '%s' CloudPrivateIPConfig", egressIPAddrStr)
+		e2e.Logf("'%s' CloudPrivateIPConfig assigned node ('%s') is kept after adding a second assignable node", egressIPAddrStr, cpicAssignedNode)
+		o.Expect(cpicAssignedNode).To(o.Equal(primaryWorker.Name), "'%s' egressIP has been deassigned to the '%s' node", egressIPAddrStr, primaryWorker.Name)
```
dulek, line 163: Do we have other tests doing some Machine operations like a reboot? I wonder if that wouldn't interfere here

test/extended/openstack/egressip.go
```diff
@@ -0,0 +1,470 @@
...
+		// Skip based on number of required worker nodes
+		g.By(fmt.Sprintf("Check the number of worker nodes (should be %d at least)", minWorkerNum))
+		e2e.Logf("Detected number of workers: '%d'", len(workerNodeList.Items))
+		if len(workerNodeList.Items) < minWorkerNum {
+			e2eskipper.Skipf("Skipping test as there are '%d' worker node(s) and minimum required is '%d'", len(workerNodeList.Items), minWorkerNum)
+		}
+
+		e2e.Logf("Getting 2 workers for the EgressIP node failover checks, aka primary and secondary workers")
+		primaryWorker := workerNodeList.Items[0]
+		secondaryWorker := workerNodeList.Items[1]
```
dulek, line 85: Do we have any tests creating Machines that will be gone later? If so we should have some way to exclude them here.

test/extended/openstack/egressip.go
```diff
@@ -0,0 +1,470 @@
...
+		e2e.Logf("Selected primary EgressIP assignable worker: %s", primaryWorker.Name)
+		e2e.Logf("Selected secondary EgressIP assignable worker: %s", secondaryWorker.Name)
+
+		// Label the primary worker node as EgressIP assignable node
+		// "dummy" is added as value for the label key just because RemoveLabelOffNode is not
+		// able to remove the key if it doesn't have any value
```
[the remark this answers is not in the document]
dulek: Okay, apparently they do that differently: https://github.com/openshift/origin/blob/1f407c1b83de99d7c05333efcdda6732bfb25dac/test/extended/networking/egressip.go#L174

dulek (changes requested):
Looks impressive, some comments inline. It's a shame the functions in https://github.com/openshift/origin/blob/6be69c609341d57364174918dd34415ef28f2172/test/extended/networking/egressip_helpers.go#L3 are private and we cannot use them.

test/extended/openstack/egressip.go
```diff
@@ -0,0 +1,470 @@
...
+		infraID = infrastructure.Status.InfrastructureName
+
+		g.By("Getting the worker node list")
+		workerNodeList, err = clientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
+			LabelSelector: "node-role.kubernetes.io/worker",
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+	})
+
+	g.It("attached to a floating IP should be kept after EgressIP node failover with OVN-Kubernetes NetworkType", func() {
```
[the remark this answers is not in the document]
rlobillo: Done

test/extended/openstack/egressip.go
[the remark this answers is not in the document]
rlobillo: we are making use of the e2e function which is including some other functionalities like retries: https://github.com/kubernetes/kubernetes/blob/v1.25.0/test/utils/density_utils.go#L37

I'll prefer to keep using it unless you think it's better to follow origin approach

test/extended/openstack/egressip.go
```diff
@@ -0,0 +1,470 @@
...
+	ipAddr := ipNetwork.IP
+	ipAddrStr := ""
+	for ; ipNetwork.Contains(ipAddr); ipAddr = incIP(ipAddr) {
+		// Skip network reserved IPs (not considering the network mask)
+		ipAddrStr = ipAddr.String()
+		splitIpAddr := strings.Split(ipAddrStr, ".")
+		var excludeIPs = []string{"0", "1", "255"}
+		if contains(excludeIPs, splitIpAddr[len(splitIpAddr)-1]) {
+			continue
+		}
```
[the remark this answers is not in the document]
rlobillo: Thank you! Updated

test/extended/openstack/egressip.go
rlobillo: afaik, we have tests creating bogus machines that never creates a node. In that case it does not affect.

rlobillo:
@dulek Thanks for your review, I've just addressed your comments.

test/extended/openstack/egressip.go
```diff
@@ -0,0 +1,475 @@
...
+
+		// Label the primary worker node as EgressIP assignable node
+		// "dummy" is added as value for the label key just because RemoveLabelOffNode is not
+		// able to remove the key if it doesn't have any value
+		g.By(fmt.Sprintf("Labeling the primary node '%s' with '%s'", primaryWorker.Name, egressAssignableLabelKey))
+		e2e.AddOrUpdateLabelOnNode(clientSet, primaryWorker.Name, egressAssignableLabelKey, "dummy")
+		defer e2e.RemoveLabelOffNode(clientSet, primaryWorker.Name, egressAssignableLabelKey)
+
+		g.By(fmt.Sprintf("Getting the EgressIP network from the '%s' annotation", egressIPConfigAnnotationKey))
+		egressIPNetCidrStr, err := getEgressIPNetwork(primaryWorker)
```
MaysaMacedo, line 94: This could be named getEgressIPAddress instead, but no strong opinion

test/extended/openstack/egressip.go
```diff
@@ -0,0 +1,475 @@
...
+		var fip *floatingips.FloatingIP
+		externalNetworkId, err := GetFloatingNetworkID(networkClient)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		fip, err = floatingips.Create(networkClient, floatingips.CreateOpts{FloatingNetworkID: externalNetworkId}).Extract()
+		o.Expect(err).NotTo(o.HaveOccurred(), "Error creating FIP using discovered floatingNetwork ID '%s'", externalNetworkId)
+		e2e.Logf("The FIP '%s' has been created in Openstack", fip.FloatingIP)
+		defer floatingips.Delete(networkClient, fip.ID)
+
+		g.By(fmt.Sprintf("Attaching the FIP %s to the EgressIP port '%s'", fip.FloatingIP, egressIPPort.Name))
+		floatingips.Update(networkClient, fip.ID, floatingips.UpdateOpts{PortID: &egressIPPort.ID})
```
MaysaMacedo, line 182: should we also .Extract() here so it returns the obj and err?
consequently we could check if there was an err and there would be no need to do lines 184-188

MaysaMacedo (commented):
It looks good, but I think we could reduce one step

test/extended/openstack/egressip.go
dulek: This is a good point!

test/extended/openstack/egressip.go
dulek: I think you're right, apparently this returns an address.

rlobillo: This function is inspecting the node and getting the CIDR for the egress Network:
```
$ oc get node/ostest-h9vbm-worker-0-fkxdr -o json | jq .metadata.annotations | grep egress
  "cloud.network.openshift.io/egress-ipconfig": "[{\"interface\":\"e4340adc-fc03-4279-8fef-a5cfe4cbf5d0\",\"ifaddr\":{\"ipv4\":\"10.196.0.0/16\"},\"capacity\":{\"ip\":8}}]",
```

I think 'network' is more accurate than 'address'...

test/extended/openstack/egressip.go
rlobillo: Done, thanks!

dulek:
/lgtm

Leaving approval to Maysa.

PR #100, opened 2023-04-18
OCPBUGS-11646: Add retries to service connectivity check

itzikb-redhat:
The following test can fail on checking connectivity to the service, Adding retries to adapt to slow setups "[sig-installer][Suite:openshift/openstack][lb][Serial] The Openstack platform should re-use an existing UDP Amphora LoadBalancer when new svc is created on Openshift with the proper annotation"

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -227,8 +227,16 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 			podNames, err := exutil.GetPodNamesByFilter(oc.KubeClient().CoreV1().Pods(oc.Namespace()),
 				exutil.ParseLabelsOrDie(""), func(v1.Pod) bool { return true })
 			o.Expect(err).NotTo(o.HaveOccurred())
+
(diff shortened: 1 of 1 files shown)


mandre:
/retitle OCPBUGS-11646: Add retries to service connectivity check

test/extended/openstack/loadbalancers.go
```diff
@@ -227,8 +227,18 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+			maxAttempts := 100
+			podName := ""
 			for _, svcPort := range [2]int32{svcPort1, svcPort2} {
-				podName, err := getPodNameThroughUdpLb(fip, fmt.Sprintf("%d", svcPort), "hostname")
+				for count := 0; count < maxAttempts; count++ {
+					podName, err = getPodNameThroughUdpLb(fip, fmt.Sprintf("%d", svcPort), "hostname")
+					if err == nil {
+						break
+					} else {
+						time.Sleep(1 * time.Second)
```
mandre, line 239: The sleep is not necessary, there's already a 5 second dealdine for the connection.

PR #104, opened 2023-05-30
OCPBUGS-13781: Don't assert identical server groups in multi-AZ

stephenfin:
In a multi-AZ deployment, there will be multiple server groups - one per AZ. Asserting that each server group is identical is therefore always going to result in a failure.

### DIFF
--- a/test/extended/openstack/servergroup.go
+++ b/test/extended/openstack/servergroup.go
@@ -34,7 +34,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 	var workerNodeList *corev1.NodeList
 	var ms dynamic.NamespaceableResourceInterface
 	var controlPlaneGroupName string
-	var workerGroupName string
+	var workerAZGroupNameMap map[string]string
 
 	oc = exutil.NewCLI("openstack")
 
@@ -85,7 +85,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 
 	// OCP 4.5: https://issues.redhat.com/browse/OSASINFRA-1300
 	g.It("creates Control plane nodes in a server group", func() {
-		g.By("Getting the the Control instances' Server groups")
+		g.By("Checking the Control plane instances are in the same Server Group")
 		{
 			for _, item := range masterNodeList.Items {
 				machineAnnotation := strings.SplitN(item.Annotations["machine.openshift.io/machine"], "/", 2)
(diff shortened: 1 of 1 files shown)


rlobillo:
@itzikb-redhat would you mind to take a look into this?

test/extended/openstack/servergroup.go
```diff
@@ -151,10 +155,14 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 
 				instanceWorkerGroupName := instanceWorkerGroupNameField.(string)
 				o.Expect(instanceWorkerGroupName).NotTo(o.BeEmpty(), "the Server group name should not be the empty string")
-				if workerGroupName == "" {
-					workerGroupName = instanceWorkerGroupName
-				} else {
-					o.Expect(instanceWorkerGroupName).To(o.Equal(workerGroupName), "two Worker Machines have different workerGroupName set")
+
+				// If there are multiple AZs, our workers will belong to separate server groups (one per AZ)
+				if availabilityZoneCount == 1 {
```
itzikb-redhat, line 160: The assumption here is that there are multiple AZs used by the cluster in case there are more than one which may be wrong (e.g. one AZ is used by the cluster , others are not)

itzikb-redhat (commented):
I suspect that in MultiAZ the following will fail too https://github.com/openshift/openstack-test/blob/9fd6cd6e783a24a44771879cd093e5c675ee813b/test/extended/openstack/servergroup.go#L169

pierreprinetti (commented):
In the main branch, we check that "**all workers are in the same server group**".

In its current form, this patch checks that "**there are as many server groups as availability zones**".

I would like to propose that we check that "**workers in each availability zone are in the same server group**". This could easily be achieved by turning `workerGroupName` from `string` to `map[string]string` where an AZ maps to a serverGroup. More importantly, I think that the name of the test should reflect that.

test/extended/openstack/servergroup.go
stephenfin: Done :+1:

stephenfin:
> In the main branch, we check that "**all workers are in the same server group**".
> 
> In its current form, this patch checks that "**there are as many server groups as availability zones**".
> 
> I would like to propose that we check that "**workers in each availability zone are in the same server group**". This could easily be achieved by turning `workerGroupName` from `string` to `map[string]string` where an AZ maps to a serverGroup. More importantly, I think that the name of the test should reflect that.

Ack, done :+1:

stephenfin:
@pierreprinetti Anything else needed here?

pierreprinetti:
/approve
/lgtm
Thank you!

PR #105, opened 2023-05-31
Move code in checking MachineSets to functions

itzikb-redhat:
We want to be able to also check masters and reuse the code

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -163,7 +163,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 						results[podName]++
 					}
 				}
-				e2e.Logf("Pods accessed after 100 UDP requests:\n%v\n", results)
+				e2e.Logf("Pods accessed after 100 requests:\n%v\n", results)
 				pods, err := oc.KubeClient().CoreV1().Pods(oc.Namespace()).List(ctx, metav1.ListOptions{})
 				o.Expect(err).NotTo(o.HaveOccurred())
 


test/extended/openstack/servers.go
```diff
@@ -152,8 +118,46 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
...
+				o.Expect(instanceImage.Name).To(o.Equal(imageMachineSet), "Image not matching for instance %q", instance.Name)
+			}
+			instanceSgs := parseInstanceSgs(instance.SecurityGroups)
+			o.Expect(instanceSgs).To(o.Equal(sgMachineSets), "SGs not matching for %q", instance.Name)
+		}
+	}
+}
+
+// listRoleMachineSets lists all machineSets based on role label
+func listRoleMachineSets(dc dynamic.Interface, role string) ([]objx.Map, error) {
```
rlobillo, line 160: What about listMachineSetsWithRole?

test/extended/openstack/servers.go
```diff
@@ -152,8 +118,46 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 	})
 })
 
-// listWorkerMachineSets lists all worker machineSets
-func listWorkerMachineSets(dc dynamic.Interface) ([]objx.Map, error) {
+// Checks that the Openstack created instances are as specified by the MachineSets
+func checkMachineSets(dc dynamic.Interface, computeClient *gophercloud.ServiceClient, machineSets []objx.Map) {
```
rlobillo, line 122: On this function you are rather checking that the machinesets'  specs are satisfied on the existing openstack instances. 

As this change is interesting for covering CPMS validations in coming patches... I am wondering if you want to check exactly the same things you are checking on the workers for the masters. I believe you'll want to do different checks. 

For example, in CPMS you should have one single machineset, so passing a list of machinesets is not applicable in that case. In fact, you should check that only one machineset should be there. Also, in the CPMS you'll have the failureDomain sections, and in worker machinesets you should not.

Therefore, I think it's cleaner to create separate checks for the CP machineset than creating a function for both worker and master machinesets.

itzikb-redhat: Maybe renaming can be beneficial. I think that we can add another function just for masters and keep a function for the common code between masters and workers

rlobillo: ok! in such a case, could you please then rename the function to something like commonCheckMachineSets and add a comment explaining that?

itzikb-redhat: yes

test/extended/openstack/servers.go
pierreprinetti: or `listMachineSetsByRole`? "role machinesets" is not very clear to me

test/extended/openstack/servers.go
```diff
@@ -182,22 +184,22 @@ func listWorkerMachineSets(ctx context.Context, dc dynamic.Interface) ([]objx.Ma
 }
 
 // getMachines lists all machines in the cluster
-func getMachines(ctx context.Context, dc dynamic.Interface) ([]objx.Map, error) {
+func getMachines(dc dynamic.Interface) ([]objx.Map, error) {
```
pierreprinetti, line 187: Why do you remove the context argument in favour of calling `context.Background()` from within the function?

test/extended/openstack/servers.go
```diff
@@ -196,7 +202,7 @@ func getMachines(ctx context.Context, dc dynamic.Interface) ([]objx.Map, error)
 }
 
 // getMachinesFromMachineSet lists all worker machines beloging to msName machineset
-func getMachinesFromMachineSet(ctx context.Context, dc dynamic.Interface, msName string) ([]objx.Map, error) {
+func getMachinesFromMachineSet(ctx context.Context,dc dynamic.Interface, msName string) ([]objx.Map, error) {
```
pierreprinetti, line 205: Please run `go fmt ./...`

test/extended/openstack/servers.go
```diff
@@ -300,13 +306,13 @@ func parseInstanceAddresses(addresses map[string]interface{}) ([]string, error)
 // If machines are not installed it skips the test case.
 // It then checks to see if the `openshift-machine-api` namespace is installed.
 // If the namespace is not present it skips the test case.
-func skipUnlessMachineAPIOperator(ctx context.Context, dc dynamic.Interface, c coreclient.NamespaceInterface) {
+func skipUnlessMachineAPIOperator(dc dynamic.Interface, c coreclient.NamespaceInterface) {
```
pierreprinetti, line 309: Please leave the `ctx` argument in

test/extended/openstack/servers.go
```diff
@@ -157,22 +122,63 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
...
+				o.Expect(instanceImage.Name).To(o.Equal(imageMachineSet), "Image not matching for instance %q", instance.Name)
+			}
+			instanceSgs := parseInstanceSgs(instance.SecurityGroups)
+			o.Expect(instanceSgs).To(o.Equal(sgMachineSets), "SGs not matching for %q", instance.Name)
+		}
+	}
+}
+
+// listRoleMachineSets lists all machineSets based on role label
+func listRoleMachineSets(dc dynamic.Interface, role string) ([]objx.Map, error) {
```
pierreprinetti, line 167: Also please leave `ctx context.Context` as the first argument

test/extended/openstack/servers.go
```diff
@@ -157,22 +122,63 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
...
 	mc := dc.Resource(schema.GroupVersionResource{
 		Group:    machineAPIGroup,
 		Version:  "v1beta1",
 		Resource: "machinesets",
 	}).Namespace(machineAPINamespace)
-	obj, err := mc.List(ctx, metav1.ListOptions{})
+	obj, err := mc.List(context.Background(), metav1.ListOptions{})
```
pierreprinetti, line 173: Please use the `ctx` received in the function signature

test/extended/openstack/servers.go
```diff
@@ -319,7 +325,7 @@ func skipUnlessMachineAPIOperator(ctx context.Context, dc dynamic.Interface, c c
 	o.Expect(err).NotTo(o.HaveOccurred())
 
 	err = wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
-		_, err := c.Get(ctx, "openshift-machine-api", metav1.GetOptions{})
+		_, err := c.Get(context.Background(), "openshift-machine-api", metav1.GetOptions{})
```
pierreprinetti, line 328: no need to use `background()` when there is `ctx`

test/extended/openstack/servers.go
```diff
@@ -46,7 +46,6 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 	g.BeforeEach(func() {
 		ctx = context.Background()
 	})
-
```
pierreprinetti, line 49: you didn't like this newline?

test/extended/openstack/machinesets.go
```diff
@@ -63,7 +63,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 
 			networkClient, err := client(serviceNetwork)
 			o.Expect(err).NotTo(o.HaveOccurred(), "Error creating an openstack network client")
-			machineSets, err := listWorkerMachineSets(ctx, dc)
+			machineSets, err := listRoleMachineSets(dc, "worker")
```
pierreprinetti, line 66: please leave `ctx` in for cancellation

test/extended/openstack/machinesets.go
```diff
@@ -54,7 +54,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 		g.It("should not have leftovers ports", func() {
 			// Check the scenario at https://bugzilla.redhat.com/show_bug.cgi?id=2073398
 
-			skipUnlessMachineAPIOperator(ctx, dc, clientSet.CoreV1().Namespaces())
+			skipUnlessMachineAPIOperator(dc, clientSet.CoreV1().Namespaces())
```
pierreprinetti, line 57: please leave `ctx` in

test/extended/openstack/servers.go
```diff
@@ -157,8 +122,49 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
...
+				o.Expect(instanceImage.Name).To(o.Equal(imageMachineSet), "Image not matching for instance %q", instance.Name)
+			}
+			instanceSgs := parseInstanceSgs(instance.SecurityGroups)
+			o.Expect(instanceSgs).To(o.Equal(sgMachineSets), "SGs not matching for %q", instance.Name)
+		}
+	}
+}
+
+// listRoleMachineSets lists all machineSets based on role label
+func listRoleMachineSets(ctx context.Context, dc dynamic.Interface, role string) ([]objx.Map, error) {
```
pierreprinetti, line 167: Sorry to be pedantic, but this name really doesn't speak to me. What do you think about `listMachineSetsByRole`?

pierreprinetti:
I understand that you want to extract checks that are likely to be common to both the ControlPlaneMachineSet and the MachineSets. Having them in a separate function might reduce code duplication when we introduce CPMS validation.

However, I suspect that this specific cut you're proposing is going to be problematic down the line.

The problematic aspect, to me, is that you are moving in one single function all the checks that may prove common to both cases, regardless of what they do. You are not bundling together checks that are logically affine: you are bundling together checks that you will use together. Typically, this becomes a problem when you need that code for a _third_ reason (after MachineSet and CPMS), where maybe only a portion of the "common" checks apply. Another reason not to conflate cases too much, is that a CPMS is not another MachineSet. It's a different resource that follows very different rules.

There is also another problem, which predates your refactoring: this test compares MachineSet properties with Nova-instance properties. It's skipping one level: we should instead compare a MachineSet with its Machines, and then (in a separate test) each Machine with its Nova-instance counterpart. Not only does this separation would make errors easier to diagnose, but it also reflects the user API. Machines can exist outside a MachineSet, and a CPMS can be disabled. These legit situations are not captured by the current test.

---

What I'm suggesting is to break the check into specific concerns. Let's start by defining the invariants that may be common across both CPMS and MachineSets:
* **MachineSet replica number corresponds to the number of Machines**
* **MachineSet ProviderSpec template is applied to Machines** (here goes the check for properties like flavor, image, root volume, secgroups & co.)

These checks go into `machinesets.go`, which might then resemble to something like this:
```Go
var _ = g.Describe("[sig-installer][Suite:openshift/openstack] MachineSet", func() {
	defer g.GinkgoRecover()

	var dc dynamic.Interface
	var clientSet *kubernetes.Clientset

	var ctx context.Context
	g.BeforeEach(func() {
		ctx = context.Background()
		skipUnlessMachineAPIOperator(ctx, dc, clientSet.CoreV1().Namespaces())
	})

	g.It("replica number corresponds to the number of Machines", func() {
		machineSets, err := listMachineSets(ctx, dc)

		for _, machineSet := range machineSets {
			machineSetName := machineSet.Get("metadata.name").String()
			machines, err := getMachinesFromMachineSet(ctx, dc, machineSetName)
			o.Expect(machines, err).To(o.HaveLen(replicasMachineSet),
				"Number of replicas not matching for machineset %q", machineSetName)
		}
	})

	g.It("ProviderSpec template is applied to Machines", func() {
		machineSets, err := listMachineSets(ctx, dc)

		for _, machineSet := range machineSets {
			machineSetName := machineSet.Get("metadata.name").String()
			machines, err := getMachinesFromMachineSet(ctx, dc, machineSetName)
			checkMachineProviderSpec(machineSet.Get("spec.template.spec.providerSpec"), machines)
		}
	})
```

Then the Machine vs Instance invariants:
* **Machine properties are reflected in OpenStack instances** (e.g. flavor, image, root volume, secgroup etc)
* **Machine status addresses contain the OpenStack instance addresses** (the other test in this file) 
* **Machine status phase matches instance status** (e.g. `Running` corresponds to the instance's status `ACTIVE`... I am not 100% sure of the mapping here)

These checks go into `servers.go`, which might then resemble to something like this:
```Go
var _ = g.Describe("[sig-installer][Suite:openshift/openstack] Machine", func() {

	g.It("properties are reflected in OpenStack instances", func() {
		for _, machine := range listMachines(ctx, dc) {
			// check that machine flavor == instance flavor
			// check that machine image or rootVolume == instance image or rootVolume
			// check that machine security gruops == instance security groups
			// etc...
		}
	})

	g.It("status addresses contain the OpenStack instance addresses", func() {}) // formerly known as "should report all openstack instance addresses on the corresponding Machine object"

	g.It("status phase matches instance status", func() {}) // which was partially implemented, but just checking "Running"


```
---

Sorry for coming up with this so late in the review process; I let myself be blinded by the removal of context propagation, which was a detail.

I am happy to pair with you on this refactoring if you're available.

test/extended/openstack/servers.go
pierreprinetti, line 167: After much thought, I think that this function should only return MachineSets. I don't even think it should bother checking the role, as it's always going to be "worker" AFAIK.

A separate function might return the CPMS.

test/extended/openstack/servers.go
```diff
@@ -157,8 +122,49 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
...
+				o.Expect(instanceImage.Name).To(o.Equal(imageMachineSet), "Image not matching for instance %q", instance.Name)
+			}
+			instanceSgs := parseInstanceSgs(instance.SecurityGroups)
+			o.Expect(instanceSgs).To(o.Equal(sgMachineSets), "SGs not matching for %q", instance.Name)
+		}
+	}
+}
+
+// listMachineSetsByRole lists all machineSets based on role label
+func listMachineSetsByRole(ctx context.Context, dc dynamic.Interface, role string) ([]objx.Map, error) {
```
pierreprinetti, line 167: (posting again because Github hid my previous comment)

This function should only deal with MachineSets. I don't even think it should bother checking the role, as it's always going to be "worker" AFAIK.

A separate function might return the CPMS.

pierreprinetti:
Superseded by https://github.com/openshift/openstack-test/pull/116

PR #106, opened 2023-06-01
Add tcp lb testcases

rlobillo:
Covering tech debt in openstack-test

https://issues.redhat.com/browse/OSASINFRA-2412

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -3,7 +3,9 @@ package openstack
 import (
 	"context"
 	"fmt"
+	"io"
 	"net"
+	"net/http"
 	"regexp"
 	"strconv"
 	"strings"
@@ -43,6 +45,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 	var cloudProviderConfig *ini.File
 	var loadBalancerServiceTimeout = 5 * time.Minute
 	availableLbProvidersUnderTests := [2]string{"Amphora", "OVN"}
+	availableProtocolsUnderTests := [2]v1.Protocol{v1.ProtocolTCP, v1.ProtocolUDP}
 	lbMethodsWithETPGlobal := map[string]string{
 		"OVN":     "source_ip_port",
 		"Amphora": "round_robin",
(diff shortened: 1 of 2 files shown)


test/extended/openstack/loadbalancers.go
```diff
@@ -84,83 +87,88 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 	for _, i := range availableLbProvidersUnderTests {
 		lbProviderUnderTest := i
 
-		// https://issues.redhat.com/browse/OSASINFRA-2753
-		g.It(fmt.Sprintf("should create an UDP %s LoadBalancer when an UDP svc with type:LoadBalancer is created on Openshift", lbProviderUnderTest), func() {
+		for _, j := range availableProtocolsUnderTests {
+			protocolUnderTest := j
```
dulek, line 91: Why like this? Just assign it directly to `protocolUnderTest`:

```
for _, protocolUnderTest := range availableProtocolsUnderTests {
```

test/extended/openstack/loadbalancers.go
```diff
@@ -601,35 +613,96 @@ func getFipbyFixedIP(client *gophercloud.ServiceClient, vip string) (floatingips
...
+
+	} else {
+		return "", fmt.Errorf("unsupported protocol %q", protocol)
+	}
+}
+
+// Send message on top of HTTP GET on provided ip and TCP port and return the answer.
 // Error if there is no answer or cannot access the port after 5 seconds.
-func getPodNameThroughUdpLb(ip string, port string, message string) (string, error) {
+func getPodNameThroughTcpLb(ip string, port string, message string) (string, error) {
```
dulek, line 682: This seems unused?

test/extended/openstack/loadbalancers.go
```diff
@@ -241,108 +249,112 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 			}
 		})
 
-		// https://issues.redhat.com/browse/OCPBUGS-2350
-		g.It(fmt.Sprintf("should apply lb-method on UDP %s LoadBalancer when an UDP svc with monitors and ETP:Local is created on Openshift", lbProviderUnderTest), func() {
+		for _, j := range availableProtocolsUnderTests {
+			protocolUnderTest := j
```
dulek, line 253: Same here.

dulek (changes requested):

dulek:
Thanks for doing this Ramon!

test/extended/openstack/loadbalancers.go
rlobillo: That's something we observed in the previous round of tests. Honestly, I don't know why but the whole engine that is calculating the list of tests don't work if I don't do that. I copied the current approach from origin, so I assume they already hit that and that was their solution.

test/extended/openstack/loadbalancers.go
rlobillo: oh right! thanks for noticing

test/extended/openstack/loadbalancers.go
pierreprinetti: right, goroutines in `for` cycles...

To make it a bit more digestible, try this:

```Go
for i := range availableProtocolsUnderTests {
			protocolUnderTest := availableProtocolsUnderTests[i]
```

pierreprinetti: I’d also check if this Ginkgo thing likes being called in a loop.

test/extended/openstack/loadbalancers.go
pierreprinetti:
```Go
for j := range availableProtocolsUnderTests {
	protocolUnderTest := availableProtocolsUnderTests[j]
```

test/extended/openstack/loadbalancers.go
```diff
@@ -87,83 +90,88 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
 
-		// https://issues.redhat.com/browse/OSASINFRA-2753
-		g.It(fmt.Sprintf("should create an UDP %s LoadBalancer when an UDP svc with type:LoadBalancer is created on Openshift", lbProviderUnderTest), func() {
+		for _, j := range availableProtocolsUnderTests {
+			protocolUnderTest := j
 
-			skipIfNotLbProvider(lbProviderUnderTest, cloudProviderConfig)
+			// https://issues.redhat.com/browse/OSASINFRA-2753
+			// https://issues.redhat.com/browse/OSASINFRA-2412
+			g.It(fmt.Sprintf("should create an %s %s LoadBalancer when an %s svc with type:LoadBalancer is created on Openshift", protocolUnderTest, lbProviderUnderTest, protocolUnderTest), func() {
```
pierreprinetti, line 98: nit: `s/when an %s svc/when a %s svc/`, which works for both UDP and TCP

test/extended/openstack/loadbalancers.go
```diff
@@ -87,83 +90,88 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
-				podName, err := getPodNameThroughUdpLb(svcIp, fmt.Sprintf("%d", svcPort), "hostname")
-				if err != nil {
-					e2e.Logf("Error detected while accessing the LoadBalancer service on try %d: %q", i, err)
-				} else {
-					results[podName]++
+				pool, err := pools.Get(loadBalancerClient, lb.Pools[0].ID).Extract()
+				o.Expect(err).NotTo(o.HaveOccurred())
+				o.Expect(pool.Protocol).Should(o.Equal(string(protocolUnderTest)), "Unexpected protocol on Openstack LoadBalancer Pool: %q", pool.Name)
+				//Set as OFFLINE in vexxhost despite the lb is operative
+				//o.Expect(pool.OperatingStatus).Should(o.Equal("ONLINE"), "Unexpected Operating Status on Openstack Pool: %q", pool.Name)
```
pierreprinetti, line 145: leftover?

test/extended/openstack/loadbalancers.go
```diff
@@ -604,35 +616,68 @@ func getFipbyFixedIP(client *gophercloud.ServiceClient, vip string) (floatingips
...
-		return "", err
+		msg := []byte(message)
+		_, err = conn.Write(msg)
+		if err != nil {
+			return "", err
+		}
+
+		// receive message from server
+		buffer := make([]byte, 1024)
+		n, _, err := conn.ReadFromUDP(buffer)
```
pierreprinetti, line 645: Since you're discarding the UPDAddr, this could be `conn.ReadFrom(buffer)`

test/extended/openstack/loadbalancers.go
```diff
@@ -604,35 +616,68 @@ func getFipbyFixedIP(client *gophercloud.ServiceClient, vip string) (floatingips
...
+			return "", err
+		}
+
+		resp, err := client.Do(req)
+		if err != nil {
+			return "", err
+		}
+		defer resp.Body.Close()
+
+		body, err := ioutil.ReadAll(resp.Body)
```
pierreprinetti, line 671: nit: use `io.ReadAll`

test/extended/openstack/loadbalancers.go
```diff
@@ -674,6 +719,9 @@ func createTestDeployment(depName string, labels map[string]string, replicas int
 	isFalse := false
 	isTrue := true
 	netExecParams := "--" + strings.ToLower(string(protocol)) + "-port=" + fmt.Sprintf("%d", port)
+	if protocol == v1.ProtocolTCP {
+		netExecParams = "--http-port=" + fmt.Sprintf("%d", port)
```
pierreprinetti, line 723: nit: this looks weird.

Either:
```Go
netExecParams = fmt.Sprintf("--http-port=%d", port)
```

or:
```Go
netExecParams = "--http-port=" + strconv.Atoi(port)
```

test/extended/openstack/loadbalancers.go
```diff
@@ -674,6 +719,9 @@ func createTestDeployment(depName string, labels map[string]string, replicas int
 	isFalse := false
 	isTrue := true
 	netExecParams := "--" + strings.ToLower(string(protocol)) + "-port=" + fmt.Sprintf("%d", port)
+	if protocol == v1.ProtocolTCP {
```
pierreprinetti, line 722: nit: to improve readability, you could treat all cases at the same time (e.g. with a case/switch) instead of setting a default and then override for TCP.

Something like:
```Go
	var netExecParams string
	switch protocol {
	case v1.ProtocolTCP:
		netExecParams = fmt.Sprintf("--http-port=%d", port)
	default:
		netExecParams = fmt.Sprintf("--%s-port=%d", strings.ToLower(string(protocol)), port)
	}
```

pierreprinetti (commented):
I'm far from being a LB expert, so this PR is probably not for me, sorry. I just added a couple comments on the form, that you can totally ignore. LGTM!

test/extended/openstack/loadbalancers.go
rlobillo: done

test/extended/openstack/loadbalancers.go
```diff
@@ -604,35 +616,68 @@ func getFipbyFixedIP(client *gophercloud.ServiceClient, vip string) (floatingips
 	return result, fmt.Errorf("FIP not found for VIP %q", vip)
 }
 
-// Send message on provided ip and UDP port and return the answer.
-// Error if there is no answer or cannot access the port after 5 seconds.
-func getPodNameThroughUdpLb(ip string, port string, message string) (string, error) {
+func getPodNameThroughLb(ip string, port string, protocol v1.Protocol, message string) (string, error) {
```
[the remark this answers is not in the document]
rlobillo: I'd prefer not do so as the port is int32 and I'm passing it converted to a string. I don't want to add another cast to convert int32 to int and then int to string. I'd let it be as it is now.

test/extended/openstack/loadbalancers.go
rlobillo: Done

test/extended/openstack/loadbalancers.go
rlobillo: <3 it. Thank you.

test/extended/openstack/loadbalancers.go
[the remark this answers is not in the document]
pierreprinetti: why is it int32?

rlobillo:
> I'm far from being a LB expert, so this PR is probably not for me, sorry. I just added a couple comments on the form, that you can totally ignore. For what it's worth, LGTM!

Thanks for your inputs!

test/extended/openstack/loadbalancers.go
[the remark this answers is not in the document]
pierreprinetti: ah `ServicePort`. Well it can be `int32` in this function signature too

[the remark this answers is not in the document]
rlobillo: in such case I cannot use strconv.Itoa(port) or equivalent strconv function, at least I wasn't able to find it.

test/extended/openstack/loadbalancers.go
rlobillo: Done

test/extended/openstack/loadbalancers.go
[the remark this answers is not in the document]
pierreprinetti: [`strconv.FormatInt(int64(i), 10)`](https://pkg.go.dev/strconv#Itoa)

dulek (approved):

dulek:
/approve

I'll leave LGTM for @MaysaMacedo.

dulek:
Whoops, my mistake, I didn't think that it'll catch that LGTM. Any review Maysa will make will need to go through a follow up patch.

test/extended/openstack/loadbalancers.go
```diff
@@ -87,83 +90,88 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+						e2e.Logf("Error detected while accessing the LoadBalancer service on try %d: %q", i, err)
+					} else {
+						results[podName]++
+					}
 				}
-			}
-			e2e.Logf("Pods accessed after 100 UDP requests:\n%v\n", results)
-			pods, err := oc.KubeClient().CoreV1().Pods(oc.Namespace()).List(ctx, metav1.ListOptions{})
-			o.Expect(err).NotTo(o.HaveOccurred())
+				e2e.Logf("Pods accessed after 100 UDP requests:\n%v\n", results)
```
MaysaMacedo, line 166: looks like the mention to UDP is a leftover here

MaysaMacedo (commented):
LGTM. I just noticed a leftover

test/extended/openstack/loadbalancers.go
rlobillo: Thanks for noticing @MaysaMacedo 
I created this PR to address this: https://github.com/openshift/openstack-test/pull/112/

pierreprinetti: hey that's a wonderful opportunity for turning `port` into a number!

PR #107, opened 2023-06-09
Add ingress-controller test

rlobillo:
https://issues.redhat.com/browse/OSASINFRA-2412

Tech debt in openstack-test: This test creates an ingress-controller matching the canary route label. It checks that the openstack loadbalancer is created with the expected attributes and access 100 times the service from outside to confirm that the functionality is provided using the expected new path (The openstack loadbalancer).

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"io"
 	"net"
@@ -20,11 +21,13 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
+	operatorv1 "github.com/openshift/api/operator/v1"
 	exutil "github.com/openshift/origin/test/extended/util"
 	ini "gopkg.in/ini.v1"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


test/extended/openstack/loadbalancers.go
```diff
@@ -553,6 +613,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 		o.Expect(successConnCount >= minSuccessConn).To(o.BeTrue(), "Found less successful connections (%d) than the minimum expected of '%d'", successConnCount, minSuccessConn)
 		e2e.Logf("Found expected number of successfull connections: '%d'", connNumber)
 	})
+
```
dulek, line 616: nit: Unneeded change?

test/extended/openstack/loadbalancers.go
```diff
@@ -412,6 +417,61 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+			svc, err := oc.AdminKubeClient().CoreV1().Services("openshift-ingress").Get(context.Background(), "router-"+name, metav1.GetOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred(), "router-%q not found in openshift-ingress namespace", name)
+			o.Expect(svc.Spec.Type).Should(o.Equal(v1.ServiceTypeLoadBalancer), "Unexpected svc Type: %q", svc.Spec.Type)
+			loadBalancerId := svc.GetAnnotations()["loadbalancer.openstack.org/load-balancer-id"]
+			o.Expect(loadBalancerId).ShouldNot(o.BeEmpty(), "load-balancer-id annotation missing")
+			o.Expect(svc.Status.LoadBalancer.Ingress).ShouldNot(o.BeEmpty(), "svc.Status.LoadBalancer.Ingress should not be empty")
+			svcIp := svc.Status.LoadBalancer.Ingress[0].IP
+			o.Expect(svcIp).ShouldNot(o.BeEmpty(), "FIP missing on svc Status")
+			o.Expect(svc.Spec.ExternalTrafficPolicy).Should(o.Equal(v1.ServiceExternalTrafficPolicyTypeLocal),
+				"Unexpected ExternalTrafficPolicy on svc specs")
```
dulek, line 446: Feels like we're testing what should be covered already in origin here, but I don't see it as an important issue.

test/extended/openstack/loadbalancers.go
```diff
@@ -412,6 +417,61 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+				poolId := lb.Pools[i].ID
+				pool, err := pools.Get(loadBalancerClient, poolId).Extract()
+				o.Expect(err).NotTo(o.HaveOccurred())
+				o.Expect(pool.Protocol).Should(o.Equal("TCP"), "Unexpected protocol on Openstack LoadBalancer Pool: %q", pool.Name)
+			}
+
+			g.By("Test that the canary service is accessible through new ingressController")
+			route, err := oc.AdminRouteClient().RouteV1().Routes("openshift-ingress-canary").Get(context.Background(), "canary", metav1.GetOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred(), "canary route not found")
+			for i := 0; i < 100; i++ {
```
dulek, line 464: I wonder if checking 100 times won't increase flakiness of the tests. It's just a question here, maybe it's just fine?

dulek:
Looks good, please fix the conflict, answer my questions inline and let's proceed with this.

test/extended/openstack/loadbalancers.go
```diff
@@ -412,6 +417,61 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+				resp, err := httpsGetWithCustomLookup(route.Spec.Host, svcIp)
```
MaysaMacedo, line 465: is only testing request for port 443 enough?

MaysaMacedo (commented):
I just have one question, but in general it LGTM.

test/extended/openstack/loadbalancers.go
rlobillo: I found issues testing the connectivty to port 80 because the app is redirecting to 443 and then I hit cert issues.

We are covering multiple pools in the test "[sig-installer][Suite:openshift/openstack][lb][Serial] The Openstack platform should re-use an existing UDP Amphora LoadBalancer when new svc is created on Openshift with the proper annotation"

test/extended/openstack/loadbalancers.go
rlobillo: This test includes also openstack checks. Origin is only openshift.

test/extended/openstack/loadbalancers.go
rlobillo: As far as I check in D/S setups and the U/S gate, it is stable. My intention here is to confirm that the traffic is correctly routed to the canary app so there is no loss of queries (as we were observing with this ETP:Local thing).

test/extended/openstack/loadbalancers.go
rlobillo: done

test/extended/openstack/loadbalancers.go
dulek: Okay, that indeed makes sense.

PR #115, opened 2023-07-13
CCM: Remove check for use-octavia option

dulek:
kubernetes/cloud-provider-openstack#1882 removed `use-octavia` option, so there's no need to set it anymore. This commit makes sure tests doesn't look for that option.

We could probably check if the option is unset, but that would create a merge cycle that would be hard to resolve, so let's just stop caring about that option.


gryf (approved):
/lgtm

PR #116, opened 2023-07-14
Refactor tests

pierreprinetti:
This patch refactors the tests targeting MachineSets, Machines and instances. This patch introduces two changes.

Change 1: invariant checking at interface boundaries.
-----------------------------------------------------

Before this patch, OpenStack instances were directly compared to MachineSets. With this patch, Machines are verified against their MachineSet, and OpenStack instances are verified against Machines.

Change 2: explicit bugfix verification
--------------------------------------

With this patch, each test that verifies a bugfix is added an explicit reference to the report, and is kept separate from the invariance checks. Bug verification must match the case it's reported with and should not be constrained by interface boundaries.

---

As a consequence, some tests have been split and renamed. Below the mapping.

**The OpenStack platform on instance creation should follow machineset specs** is now split into:
* `MachineSet have role worker`
* `MachineSet replica number corresponds to the number of Machines`
* `MachineSet ProviderSpec template is correctly applied to Machines`
* `Machine are in phase Running`
* `Machine ProviderSpec is correctly applied to OpenStack instances`

**The OpenStack platform on instance creation should report all openstack instance addresses on the corresponding Machine object** is now:
* `Bugfix bz_2022627: Machine should report all openstack instance addresses`

**The OpenStack platform after deletion of a machineset should not have leftovers ports** is now:
* `Bugfix bz_2073398: MachineSet scale-in does not leak OpenStack ports`

---

## Notes for the reviewer

The reviewer(s) should answer these questions:
1. Is the pattern OK?
2. Did I correctly refactor all of the assertions that were in place, or did I skip any?


pierreprinetti:
If you foks agree with the pattern, we can complete the refactoring with the other bugfixes. Having `bz_12345` or `OSASINFRA-XXXX` in the test name will allow us to automatically link verifications to Jira cards in the future.

pierreprinetti:
Two tests interfered. That's something we should tackle, but not in this PR
/retest

pierreprinetti:
>Two tests interfered. That's something we should tackle, but not in this PR

OK, I take that back. Marked as `[Serial]`

PR #117, opened 2023-07-28
Read processed cloud.conf from ConfigMap

dulek:
Tests were reading the openshift-config/cloud-provider-config ConfigMap to get the current cloud provider settings. This is the ConfigMap that te user can use to set his own settings. There are however also settings that CCMO injects and the actual cloud.conf is produced by combining these with user input.

This commit makes sure we read the ConfigMap at
openshift-cloud-controller-manager/cloud-conf which is the one actually fed to cloud-provider-openstack. This way tests can take OpenShift defaults into account.


stephenfin:
Yeah, this is the correct config map. Good spot.

/lgtm
/approve

dulek:
/retest

Invalid checksum, not related.

PR #120, opened 2023-09-06
Bump LB creation timeout to 10 minutes

dulek:
Current limit of 5 minutes seems to be way too low for Vexxhost running with OVN. This commit bumps the timeout to 10 minutes.


dulek:
/override ci/prow/test-kuryr

We don't run these tests with Kuryr. Also Kuryr job is broken without #1988.

PR #121, opened 2023-09-06
OCPBUGS-18607: skip LB tests when cluster is behind a proxy

EmilienM:
We have observed that LB tests don't behave well when the OCP cluster
runs behind a proxy, like it's the case for most of the clusters running
in our Mecha and NFV clouds.

We have coverage of these tests on other clouds (VH managed), so we can
skip these tests for now in this scenario.

In the future, we want to improve that and make it so some LB tests can
be run behind a proxy (e.g. the TCP ones).

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"os"
 	"regexp"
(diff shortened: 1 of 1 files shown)


dulek:
/approve

I'd like @eurijon or @rlobillo to take a look here too.

mandre (commented):
/lgtm

PR #122, opened 2023-09-07
OCPBUGS-18607: re-add proxy.Status.HTTPProxy check

EmilienM:
If a cluster is running in a disconnected environment (ie BYON), the LBs
can't be created because right now the tests rely on Floating-IPs, which
is not a resource we can use in a restricted network.

Re-adding the previous condition, aside of the new one to cover both
use-cases.


stephenfin:
I was going to suggest creating a revert, but we're actually re-adding the original check *in addition to* the new one @EmilienM replaced this with in 9d74f3447. Makes sense.

/lgtm

PR #124, opened 2023-10-17
Verify machines attributes conform to CPMS

itzikb-redhat:
Checking
 - Flavor
 - Image
 - Security Groups
 - Subnets
 - Nova Availability Zone

PR #126, opened 2023-10-26
OSASINFRA-3278: add tests when etcd runs on additional local disk

EmilienM:
If both conditions are met:
* Machine API operator is deployed and managing the Machines
* a CPMS exists with an additional block device named `etcd`

Then we will check that:
* the Control Plane flavor in Nova has enough Ephemeral disk for storing etcd data
* all the Control Plane Machines have an ephemeral disk for etcd, of type `Local` and with the minimum size required

### DIFF
--- a/test/extended/openstack/local-bd-etcd.go
+++ b/test/extended/openstack/local-bd-etcd.go
@@ -1,0 +1,139 @@
+package openstack
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	"github.com/openshift/openstack-test/test/extended/openstack/machines"
+	"github.com/stretchr/objx"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+)
+
+var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenShift cluster", func() {
+	defer g.GinkgoRecover()
+
+	const minEtcdDiskSizeGiB = 7
+	const etcdBlockDeviceName = "etcd"
+
+	var computeClient *gophercloud.ServiceClient
+	var ctx context.Context
+	var dc dynamic.Interface
+	var controlPlaneFlavor string
+	var clusterCPMS objx.Map
+	var clientSet *kubernetes.Clientset
+
+	checkEtcdDisk := func(machine objx.Map) error {
+		additionalBlockDevices := machine.Get("spec.providerSpec.value.additionalBlockDevices").ObjxMapSlice()
+		if len(additionalBlockDevices) == 0 {
+			return fmt.Errorf("machine %q does not have additional block devices", machine.Get("metadata.name").String())
+		}
+
+		for _, blockDevice := range additionalBlockDevices {
+			if blockDevice.Get("name").String() == etcdBlockDeviceName {
+				if blockDevice.Get("sizeGiB").Int() < minEtcdDiskSizeGiB {
+					return fmt.Errorf("machine %q has an additional block device with size %d GiB, which is less than the minimum required %d GiB", machine.Get("metadata.name").String(), blockDevice.Get("sizeGiB").Int(), minEtcdDiskSizeGiB)
+				}
+				if blockDevice.Get("storage.type").String() != "Local" {
+					return fmt.Errorf("machine %q has an additional block device with storage type %q, which is not Local", machine.Get("metadata.name").String(), blockDevice.Get("storage.type").String())
+				}
+				return nil
+			}
+		}
+		return fmt.Errorf("machine %q does not have an additional block device named %s", machine.Get("metadata.name").String(), etcdBlockDeviceName)
+	}
+
+	getControlPlaneMachineSets := func(ctx context.Context, dc dynamic.Interface) ([]objx.Map, error) {
+		mc := dc.Resource(schema.GroupVersionResource{
+			Group:    machineAPIGroup,
+			Version:  "v1",
+			Resource: "controlplanemachinesets",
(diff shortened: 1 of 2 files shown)


EmilienM:
```
: [sig-installer][Suite:openshift/openstack] The OpenStack platform checks resources when etcd is on an ephemeral block device

Reason: skip [github.com/openshift/openstack-test/test/extended/openstack/local-bd-etcd.go:151]: CPMS does not have additional block devices Ginkgo exit error 3: exit with code 3
```

This is now skipped as expected.

I'm going to add a new CI job to test it, I'm not sure what we want first : the test or the job?

test/extended/openstack/local-bd-etcd.go
```diff
@@ -0,0 +1,162 @@
...
+	"k8s.io/client-go/kubernetes"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+)
+
+const minEtcdDiskSizeGiB = 7
+const etcdBlockDeviceName = "etcd"
+
+var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack platform", func() {
+	defer g.GinkgoRecover()
```
pierreprinetti, line 24: Please keep variables local to their context:

```suggestion
var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack platform", func() {
	defer g.GinkgoRecover()

	const minEtcdDiskSizeGiB = 7
	const etcdBlockDeviceName = "etcd"
```

test/extended/openstack/local-bd-etcd.go
```diff
@@ -0,0 +1,162 @@
...
+
+		computeClient, err = client(serviceCompute)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		skipUnlessMachineAPIOperator(ctx, dc, clientSet.CoreV1().Namespaces())
+		skipUnlessEtcdAdditionalBlockDevice(ctx, dc)
+
+	})
+
+	g.It("checks resources when etcd is on an ephemeral block device", func() {
```
pierreprinetti, line 52: This will render as:
> "[sig-installer][Suite:openshift/openstack] The OpenStack platform checks resources when etcd is on an ephemeral block device

which makes little sense. Can you please rephrase in terms of what you expect to be in the cluster?

test/extended/openstack/local-bd-etcd.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, m := range objects(objx.Map(obj.UnstructuredContent()).Get("items")) {
+		if m.Get("machine.openshift.io/cluster-api-machine-role").String() == "master" {
+			controlPlaneMachines = append(controlPlaneMachines, m)
+		}
+	}
+	return controlPlaneMachines, nil
+}
+
+// checkEtcdDisk checks that the machine has an additional block device named etcd with the correct size and storage type.
+func checkEtcdDisk(machine objx.Map) error {
```
pierreprinetti, line 105: If this function is only used in this test, consider moving it in the context of the test itself:

```Go
var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack platform", func() {
	defer g.GinkgoRecover()
	checkEtcdDisk := func(machine obj.Map) error { 
```

test/extended/openstack/local-bd-etcd.go
```diff
@@ -0,0 +1,162 @@
...
+			for _, cpMachine := range controlPlaneMachines {
+				o.Expect(checkEtcdDisk(cpMachine)).To(o.Succeed())
+			}
+		}
+
+	})
+})
+
+// getControlPlaneMachines returns the control plane machines.
+func getControlPlaneMachines(ctx context.Context, dc dynamic.Interface) ([]objx.Map, error) {
```
pierreprinetti, line 85: FYI I am experimenting a convenient API for getting machines (including by role, like in this case) here https://github.com/openshift/openstack-test/pull/127

test/extended/openstack/local-bd-etcd.go
EmilienM: No because they are used in functions too at the end of the file and if I move them, it doesn't compile.

test/extended/openstack/local-bd-etcd.go
EmilienM: Really cool and useful, I'll rebase on top of your PR if that merges before or I'll come back to it if mine merges before.

test/extended/openstack/local-bd-etcd.go
EmilienM: FYI I moved `getControlPlaneMachines` too.

test/extended/openstack/local-bd-etcd.go
```diff
@@ -0,0 +1,158 @@
...
+			Group:    machineAPIGroup,
+			Version:  "v1",
+			Resource: "machines",
+		}).Namespace(machineAPINamespace)
+		obj, err := mc.List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return nil, err
+		}
+		for _, m := range objects(objx.Map(obj.UnstructuredContent()).Get("items")) {
+			if m.Get("machine.openshift.io/cluster-api-machine-role").String() == "master" {
```
pierreprinetti, line 66: I am not sure this is correct. At a minimum, shouldn't you look into a `labels` property?

Anyway, you can start using https://github.com/openshift/openstack-test/pull/127 as soon as we merge it (and if I got it wrong, then we will collectively notice & fix!)

EmilienM: yeah I'll refactor on your PR, yours will merge today I think. I'm waiting for CI and I'll approve.
Thanks for your work!

pierreprinetti:
/lgtm
Under the (tested) assumption that this doesn’t bring regressions to the existing jobs, and as a base for the new job.

pierreprinetti (approved):

EmilienM:
NOTE: this doesn't work. I tested locally and the test is skipped because it can't find the CPMS data. I'll fix it in another PR once https://github.com/openshift/release/pull/44991 has landed.

pierreprinetti:
@itzikb-redhat has some code that pulls the CPMS here: https://github.com/openshift/openstack-test/pull/124

test/extended/openstack/local-bd-etcd.go
mandre: #127 has merged, should we add the changes Pierre asked for in https://github.com/openshift/openstack-test/pull/128 ?

EmilienM: that's what I did in the other PR? I think your comment it outdated

PR #128, opened 2023-10-28
local-bd-etcd: fix CPMS

EmilienM:
* Re-use code from cpms.go
* Simplify error handling in cpms.go for errors out of getCPMS.
  It wasn't useful to count the number of CPMS, it'll always be zero
  when an error occurs.
* If an error is found when listing CPMS, just skip the test.
  We don't care how many CPMS are found, there is already a check
  in the function. This test can only function when one CPMS is found. OpenShift CPMS can only have 1 CPMS for now.
* Stop returning numCpms, we don't need to know at it should always
  be one, otherwise we return an error.
* Increase the size of the block device required for etcd. etcd grow to
  max 8GB, so let's have some window and require 10 minimum.

### DIFF
--- a/test/extended/openstack/cpms.go
+++ b/test/extended/openstack/cpms.go
@@ -27,7 +27,6 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
 	var clientSet *kubernetes.Clientset
 	var controlPlaneMachineSet objx.Map
 	var controlPlaneMachines []objx.Map
-	var numCpms int
 
 	g.BeforeEach(func() {
 		ctx = context.Background()
@@ -41,14 +40,9 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		skipUnlessMachineAPIOperator(ctx, dc, clientSet.CoreV1().Namespaces())
-		controlPlaneMachineSet, numCpms, err = getControlPlaneMachineSet(ctx, dc)
+		controlPlaneMachineSet, err = getControlPlaneMachineSet(ctx, dc)
 		if err != nil {
-			if numCpms == 0 {
-				// In some cases (e.g. UPI), the control plane machine set is not created.
-				e2eskipper.Skipf("Failed to get control plane machine sets: %v", err)
-			} else {
-				o.Expect(err).NotTo(o.HaveOccurred())
-			}
+			e2eskipper.Skipf("Failed to get control plane machine sets: %v", err)
 		}
 		controlPlaneMachines, err = machines.List(ctx, dc, machines.ByRole("master"))
 		o.Expect(err).NotTo(o.HaveOccurred())
@@ -129,8 +123,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
 	})
 })
 
-func getControlPlaneMachineSet(ctx context.Context, dc dynamic.Interface) (objx.Map, int, error) {
-	var numCpms int = 0
+func getControlPlaneMachineSet(ctx context.Context, dc dynamic.Interface) (objx.Map, error) {
 	mc := dc.Resource(schema.GroupVersionResource{
 		Group:    machineAPIGroup,
 		Version:  "v1",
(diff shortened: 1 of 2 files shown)


test/extended/openstack/local-bd-etcd.go
```diff
@@ -28,9 +28,33 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenShift clu
 	var ctx context.Context
 	var dc dynamic.Interface
 	var controlPlaneFlavor string
-	var clusterCPMS objx.Map
 	var clientSet *kubernetes.Clientset
 
+	getControlPlaneMachineSets := func(ctx context.Context, dc dynamic.Interface) ([]objx.Map, error) {
```
itzikb-redhat, line 33: I have getControlPlaneMAchineSet at cpms.go https://github.com/openshift/openstack-test/pull/124/commits/cb55d18920ac8cd2724122e427f17709d6d45f35#diff-7b12ea0dbd387e5b0d5a4eef6d705422cd7eabd7cd8093be463d17d9eb882841R132 , Please take a look and tell me if you think it should be removed/moved etc

EmilienM: excellent, I'll refactor :-)

test/extended/openstack/cpms.go
```diff
@@ -43,13 +43,9 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
 		skipUnlessMachineAPIOperator(ctx, dc, clientSet.CoreV1().Namespaces())
 		controlPlaneMachineSet, numCpms, err = getControlPlaneMachineSet(ctx, dc)
 		if err != nil {
-			if numCpms == 0 {
```
EmilienM, line 46: this was not useful. in case of an error, `numCpms` will always be `0`. Let's just skip the test if CPMS can't be found (in case of UPI for example).

test/extended/openstack/cpms.go
```diff
@@ -43,13 +43,9 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
...
-				// In some cases (e.g. UPI), the control plane machine set is not created.
-				e2eskipper.Skipf("Failed to get control plane machine sets: %v", err)
-			} else {
-				o.Expect(err).NotTo(o.HaveOccurred())
-			}
+			e2eskipper.Skipf("Failed to get control plane machine sets: %v", err)
```
MaysaMacedo, line 46: If any error happen we would skip the test, is this really desired?

EmilienM: yes, we rely on CPMS data, so let's skip the test if we can't get it.

MaysaMacedo: disregard my comment. I didn't see your last comment before I commented

MaysaMacedo:
/lgtm

let's see result of jobs

MaysaMacedo:
/lgtm
wait for CI before approval

MaysaMacedo:
/approve

All jobs are required anyway

EmilienM:
@pierreprinetti @itzikb-redhat See the change in `cpms.go`; I think I simplified it a bit. Also note that I made the change to cpms.go into a separate commit so this can be cherry-picked without the other file, since the etcd tests is only for 4.15 and beyond.

@MaysaMacedo I think I've fixed the error we discussed offline, we'll see how CI works now.

EmilienM:
I tested locally with a deployment with etcd on additional disk and the test was skipped so something is wrong...

I'll look into it tomorrow. Sorry for the UPI job breakage in the meantime.

EmilienM:
```
passed: (6.5s) 2023-10-31T01:10:19 "[sig-installer][Suite:openshift/openstack] The OpenShift cluster runs with etcd on ephemeral local block device"
```
Got it working :-)

test/extended/openstack/cpms.go
itzikb-redhat: Is it possible that due to a bug there will be more than one CPMS?

test/extended/openstack/local-bd-etcd.go
```diff
@@ -102,15 +78,20 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenShift clu
...
+
+		skipUnlessEtcdAdditionalBlockDevice(cpms)
 	})
 
 	g.It("runs with etcd on ephemeral local block device", func() {
 
 		g.By("checking that the control plane flavor has enough ephemeral storage")
 		{
-			controlPlaneFlavor = clusterCPMS.Get("spec.template.spec.providerSpec.value.flavor").String()
+			controlPlaneFlavor = cpms.Get("spec.template.machines_v1beta1_machine_openshift_io.spec.providerSpec.value.flavor").String()
```
itzikb-redhat, line 94: You may want to declare  the provider spec value spec.template.machines_v1beta1_machine_openshift_io.spec.providerSpec.value once

EmilienM: done

EmilienM:
Tested with latest patchset:
```
passed: (5.9s) 2023-10-31T13:02:22 "[sig-installer][Suite:openshift/openstack] The OpenShift cluster runs with etcd on ephemeral local block device"
```

EmilienM:
/hold cancel
I think this is good to go, we can always iterate later.
@itzikb-redhat I'll let you cherry-pick the second commit to 4.14 if you want.

MaysaMacedo (approved):

PR #132, opened 2023-10-31
Check the Machine Failure domains conform with CPMS

itzikb-redhat:
Check:
- Nova availability zones
- Root volume Cinder availability zones and volume types
- Additional networks

### DIFF
--- a/test/extended/openstack/cpms.go
+++ b/test/extended/openstack/cpms.go
@@ -3,7 +3,6 @@ package openstack
 import (
 	"context"
 	"fmt"
-	"sort"
 	"strconv"
 
 	g "github.com/onsi/ginkgo/v2"
@@ -65,8 +64,22 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
 		cpmsSecurityGroups := make(map[string]struct{})
 		cpmsServerGroup := providerSpec.Get("value.serverGroupName").String()
 		cpmsSubnets := make(map[string]struct{})
-		var machineNovaAvailabilityZones []string
-		var cpmsNovaAvailabilityZones []string
+		var cpmsNetworks []string
+		type FailureDomain struct {
+			ComputeZone string
+			VolumeZone  string
+			VolumeType  string
+		}
+		var cpmsFDs []FailureDomain
+		var machinesFDs = make(map[FailureDomain]struct{})
+
+		for _, network := range objects(providerSpec.Get("value.networks")) {
+			subnet := network.Get("subnets").String()
+
+			if subnet == "" {
+				cpmsNetworks = append(cpmsNetworks, network.Get("uuid").String())
+			}
+		}
 
 		for _, sg := range objects(providerSpec.Get("value.securityGroups")) {
 			securityGroupName := sg.Get("name").String()
@@ -81,26 +94,29 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
 
 		if fds.String() != "" {
 			for _, fd := range objects(controlPlaneMachineSet.Get("spec.template.machines_v1beta1_machine_openshift_io.failureDomains.openstack")) {
-				nova_az := fd.Get("availabilityZone").String()
-				if nova_az != "" {
-					cpmsNovaAvailabilityZones = append(cpmsNovaAvailabilityZones, nova_az)
-				}
+				cpmsFDs = append(cpmsFDs, FailureDomain{
+					ComputeZone: fd.Get("availabilityZone").String(),
+					VolumeZone:  fd.Get("rootVolume.availabilityZone").String(),
+					VolumeType:  fd.Get("rootVolume.volumeType").String(),
+				})
 			}
 		}
-		sort.Strings(cpmsNovaAvailabilityZones)
-
 		for _, machine := range controlPlaneMachines {
 			machineName := machine.Get("metadata.name").String()
 			g.By("Comparing the MachineSet spec with machine " + machineName)
 
+			machineProviderSpec := machine.Get("spec.providerSpec.value").ObjxMap()
 			machineFlavor := machine.Get("spec.providerSpec.value.flavor").String()
 			machineImage := machine.Get("spec.providerSpec.value.image").String()
 			machineSecurityGroups := make(map[string]struct{})
 			machineSubnets := make(map[string]struct{})
-			machineServerGroup := machine.Get("spec.providerSpec.value.serverGroupName").String()
-			machineNovaAvailabilityZone := machine.Get("spec.providerSpec.value.availabilityZone").String()
+			machineServerGroup := machineProviderSpec.Get("serverGroupName").String()
+			machineNovaAz := machine.Get("spec.providerSpec.value.availabilityZone").String()
+			machineCinderAz := machine.Get("spec.providerSpec.value.rootVolume.availabilityZone").String()
+			machineVolumeType := machine.Get("spec.providerSpec.value.rootVolume.volumeType").String()
+			var machineNetworks []string
 
-			for _, sg := range objects(machine.Get("spec.providerSpec.value.securityGroups")) {
+			for _, sg := range objects(machineProviderSpec.Get("securityGroups")) {
 				securityGroupName := sg.Get("name").String()
 				machineSecurityGroups[securityGroupName] = struct{}{}
 			}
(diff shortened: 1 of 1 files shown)


test/extended/openstack/cpms.go
```diff
@@ -81,45 +91,87 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
...
 			o.Expect(machineFlavor).To(o.Equal(cpmsFlavor), "flavor mismatch on Machine %q", machineName)
 			o.Expect(machineImage).To(o.Equal(cpmsImage), "image mismatch on Machine %q", machineName)
 			o.Expect(machineSecurityGroups).To(o.Equal(cpmsSecurityGroups), "security group mismatch on Machine %q", machineName)
 			o.Expect(machineSubnets).To(o.Equal(cpmsSubnets), "subnets mismatch on Machine %q", machineName)
 			o.Expect(machineServerGroup).To(o.Equal(cpmsServerGroup), "server group	 mismatch on Machine %q", machineName)
+			o.Expect(machineNetworks).To(o.Equal(cpmsNetworks), "Network mismatch on Machine %q", machineName)
 		}
-		sort.Strings(machineNovaAvailabilityZones)
-		o.Expect(machineNovaAvailabilityZones).To(o.Equal(cpmsNovaAvailabilityZones))
+		o.Expect(sortFd(machinesFds)).To(o.Equal(sortFd(cpmsFds)))
```
pierreprinetti, line 174: Luckily, Gomega [has a matcher](https://github.com/itzikb-redhat/openstack-test/blob/18c279561a659ede4a11afc53ce5f4b25465a151/vendor/github.com/onsi/gomega/matchers.go#L349) that doesn't care about ordering. You should be able to drop the sorting logic:

```suggestion
		o.Expect(machinesFDs).To(o.ConsistOf(cpmsFDs))
```

test/extended/openstack/cpms.go
```diff
@@ -81,45 +91,87 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
...
+					fdMap["nova_az"] = novaAz
+				}
+				if cinderAz != "" {
+					fdMap["cinder_az"] = cinderAz
+				}
+				if volumeType != "" {
+					fdMap["volume_type"] = volumeType
 				}
+				cpmsFds = append(cpmsFds, fdMap)
+			}
```
pierreprinetti, line 111: Consider defining a local FailureDomain type above:

```Go
		type FailureDomain struct {
			ComputeZone string
			VolumeZone  string
			VolumeType  string
		}

		var cpmsFDs []FailureDomain
		var machinesFDs []FailureDomain
```

Then you may be able to compress this whole block into:

```suggestion
				cpmsFDs = append(cpmsFDs, FailureDomain{
					ComputeZone: fd.Get("availabilityZone").String(),
					VolumeZone:  fd.Get("rootVolume.availabilityZone").String(),
					VolumeType:  fd.Get("rootVolume.volumeType").String(),
				})
```

test/extended/openstack/cpms.go
```diff
@@ -65,8 +65,18 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
 		cpmsSecurityGroups := make(map[string]struct{})
 		cpmsServerGroup := providerSpec.Get("value.serverGroupName").String()
 		cpmsSubnets := make(map[string]struct{})
-		var machineNovaAvailabilityZones []string
+		var cpmsNetworks []string
 		var cpmsNovaAvailabilityZones []string
+		cpmsFds := make([]map[string]string, 0)
+		machinesFds := make([]map[string]string, 0)
```
pierreprinetti, line 71: Consider using slices of a custom type instead, to leverage strong typing (see my suggestion about `type FailureDomain` below)

pierreprinetti (commented):
I have a couple comments around these points:
* you assume that the given CPMS has either zero, one or three failure domains. What if there are two? (I might be remembering validation incorrectly)
* sorting should not be necessary if you use a specific Gomega matcher

Also, the variable `cpmsNovaAvailabilityZones` appears unused now. Do you think we can remove it?

test/extended/openstack/cpms.go
```diff
@@ -81,45 +91,87 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
...
+				}
+				if volumeType != "" {
+					fdMap["volume_type"] = volumeType
 				}
+				cpmsFds = append(cpmsFds, fdMap)
+			}
+			// In case we have a single Fds we replicate it so that we can compare it to the list of Fds
+			// from the machines
+			if len(cpmsFds) == 1 {
+				duplicateFd(&cpmsFds)
```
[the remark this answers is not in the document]
itzikb-redhat: @pierreprinetti If we have 2 FDs I think it will fail unless we have 2 machines

test/extended/openstack/cpms.go
```diff
@@ -139,3 +169,14 @@ func getControlPlaneMachineSet(ctx context.Context, dc dynamic.Interface) (objx.
 	}
 	return cpmsList[0], nil
 }
+
+type FailureDomain struct {
```
pierreprinetti, line 173: Please define this type locally to the test function, as it is not going to be useful elsewhere

test/extended/openstack/cpms.go
```diff
@@ -139,3 +169,14 @@ func getControlPlaneMachineSet(ctx context.Context, dc dynamic.Interface) (objx.
...
+	ComputeZone string
+	VolumeZone  string
+	VolumeType  string
+}
+
+// Duplicate a FD
+func duplicateFd(fdList *[]FailureDomain) {
```
pierreprinetti, line 180: Please define this function locally to the test function

```Go
duplicateFD := func(fdList *[]FailureDomain) {
```

test/extended/openstack/cpms.go
```diff
@@ -109,17 +126,30 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
 				subnetName := subnet.Get("filter.name").String()
 				machineSubnets[subnetName] = struct{}{}
 			}
-			if machineNovaAvailabilityZone != "" {
-				machineNovaAvailabilityZones = append(machineNovaAvailabilityZones, machineNovaAvailabilityZone)
+			if machineNovaAz != "" || machineCinderAz != "" || machineVolumeType != "" {
```
pierreprinetti, line 129: You don't need these `!= ""` checks, as the struct properties will be the empty string anyway

test/extended/openstack/cpms.go
```diff
@@ -109,17 +126,30 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
...
+					VolumeType:  machineVolumeType,
+				})
 			}
+
+			for _, network := range objects(machine.Get("spec.providerSpec.value.networks")) {
+				subnet := network.Get("subnets").String()
+
+				if subnet == "" {
+					machineNetworks = append(machineNetworks, network["uuid"].(string))
+				}
```
pierreprinetti, line 142: You don't use that `subnet` variable, am I right?

```suggestion
				if network.Get("subnets").String() == "" {
					machineNetworks = append(machineNetworks, network.Get("uuid").String())
				}
```

test/extended/openstack/cpms.go
```diff
@@ -109,17 +126,30 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
...
+			}
+
 			o.Expect(machineFlavor).To(o.Equal(cpmsFlavor), "flavor mismatch on Machine %q", machineName)
 			o.Expect(machineImage).To(o.Equal(cpmsImage), "image mismatch on Machine %q", machineName)
 			o.Expect(machineSecurityGroups).To(o.Equal(cpmsSecurityGroups), "security group mismatch on Machine %q", machineName)
 			o.Expect(machineSubnets).To(o.Equal(cpmsSubnets), "subnets mismatch on Machine %q", machineName)
 			o.Expect(machineServerGroup).To(o.Equal(cpmsServerGroup), "server group	 mismatch on Machine %q", machineName)
+			o.Expect(machineNetworks).To(o.Equal(cpmsNetworks), "Network mismatch on Machine %q", machineName)
```
pierreprinetti, line 150: Do you think it makes sense to use `o.ConsistOf` here as well?

test/extended/openstack/cpms.go
itzikb-redhat: I'm replacing it with the line of code instead

test/extended/openstack/cpms.go
pierreprinetti: Sorted out out-of-band: I now think we need this check

test/extended/openstack/cpms.go
itzikb-redhat: right. Updated

test/extended/openstack/cpms.go
itzikb-redhat: We care about the order here , right? And o.consistOf doesn't care about it (right?)

pierreprinetti: I propose ConsistOf specifically because it doesn't care about order, and I mistakenly thought we also didn't.

pierreprinetti:
/approve
/lgtm

unless the test says no.

PR #136, opened 2023-11-02
test that internal-lb cannot be shared with external-lb

rlobillo:
As metioned on k8s doc[1]:

In order to prevent accidental exposure internal Services cannot share a load balancer with any other Service. This means that cloud provider will prevent creation of a secondary internal Service sharing a load balancer with either external or internal Service. This is because floating IPs are attached to the load balancer and not to the listener.

This test is for covering that scenario.

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#sharing-load-balancer-with-multiple-services

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -28,6 +28,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/uuid"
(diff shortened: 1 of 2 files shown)


test/extended/openstack/loadbalancers.go
```diff
@@ -258,6 +260,77 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
+			o.Expect(err).To(o.HaveOccurred()) //Error is expected
+			o.Expect(svc2).To(o.BeNil())
+			g.By("Waiting for event reporting the error creating the internal lb")
+			eventSelector := fields.Set{
+				"involvedObject.kind":      "Service",
+				"involvedObject.name":      svcName2,
+				"involvedObject.namespace": oc.Namespace(),
+				"reason":                   "SyncLoadBalancerFailed",
+			}.AsSelector().String()
+			msg := "Error syncing load balancer: failed to ensure load balancer: internal Service cannot share a load balancer"
```
dulek, line 322: Oh wow, I don't think we should depend on the message too much, any change there (like adapting to a new logging format) would invalidate the test for us. Moreover it'll be increasingly hard to keep it up to date if message changes in some CPO version.

Can we at a minimum only look for the "internal Service cannot share a load balancer" part?

dulek (commented):
Looking good, just one remark - let's make sure we don't depend on the exact wording of the event. It's not something that's part of the API contract and this can change fairly easily.

test/extended/openstack/loadbalancers.go
rlobillo: Right! I have updated to look for "internal Service cannot share a load balancer" . Thanks!

PR #144, opened 2023-11-30
Adapt tests to ipv6primary dualstack cluster

rlobillo:
[KURYRQE-1414](https://issues.redhat.com//browse/KURYRQE-1414)

While running openstack-test in ipv6primary dualstack cluster, there are some tests that are not passing.

This patch performs the needed actions to make those tests available in both ipv4 and ipv6 primary dualstack cluster.

### DIFF
--- a/test/extended/openstack/egressip.go
+++ b/test/extended/openstack/egressip.go
@@ -73,6 +73,12 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][egressip] An egre
 
 	g.It("attached to a floating IP should be kept after EgressIP node failover with OVN-Kubernetes NetworkType", func() {
 
+		dualstackIpv6Primary, err := isIpv6primaryDualStackCluster(ctx, oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if dualstackIpv6Primary { //This test is covering and scenario that has no sense with ipv6 as there is no FIP/egressIP association.
+			e2eskipper.Skipf("Test not applicable for ipv6primary dualstack environments")
+		}
+
 		g.By("Getting the network type")
 		networkType, err := getNetworkType(ctx, oc)
 		o.Expect(err).NotTo(o.HaveOccurred())
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -132,7 +132,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 				o.Expect(loadBalancerId).ShouldNot(o.BeEmpty(), "load-balancer-id annotation missing")
 				o.Expect(svc.Status.LoadBalancer.Ingress).ShouldNot(o.BeEmpty(), "svc.Status.LoadBalancer.Ingress should not be empty")
 				svcIp := svc.Status.LoadBalancer.Ingress[0].IP
-				o.Expect(svcIp).ShouldNot(o.BeEmpty(), "FIP missing on svc Status")
+				o.Expect(svcIp).ShouldNot(o.BeEmpty(), "Ingress IP missing on svc Status")
 				o.Expect(svc.Spec.ExternalTrafficPolicy).Should(o.Equal(v1.ServiceExternalTrafficPolicyTypeCluster),
 					"Unexpected ExternalTrafficPolicy on svc specs")
 
@@ -140,10 +140,11 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 				lb, err := octavialoadbalancers.Get(loadBalancerClient, loadBalancerId).Extract()
 				o.Expect(err).NotTo(o.HaveOccurred())
 				o.Expect(lb.Provider).Should(o.Equal(strings.ToLower(lbProviderUnderTest)), "Unexpected provider in the Openstack LoadBalancer")
-
-				fip, err := getFipbyFixedIP(networkClient, lb.VipAddress)
-				o.Expect(err).NotTo(o.HaveOccurred())
-				o.Expect(fip.FloatingIP).Should(o.Equal(svcIp), "Unexpected floatingIp in the Openstack LoadBalancer")
+				if isIpv4(lb.VipAddress) { // No FIP assignment on ipv6
+					fip, err := getFipbyFixedIP(networkClient, lb.VipAddress)
+					o.Expect(err).NotTo(o.HaveOccurred())
+					o.Expect(fip.FloatingIP).Should(o.Equal(svcIp), "Unexpected floatingIp in the Openstack LoadBalancer")
+				}
 				o.Expect(lb.Pools).Should(o.HaveLen(1), "Unexpected number of pools on Openstack LoadBalancer %q", lb.Name)
 
 				pool, err := pools.Get(loadBalancerClient, lb.Pools[0].ID).Extract()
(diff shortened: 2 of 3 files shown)


test/extended/openstack/utils.go
```diff
@@ -332,3 +384,29 @@ func conditionsMatchExpected(expected, actual map[string]string) bool {
 	}
 	return reflect.DeepEqual(expected, filtered)
 }
+
+// IsIPv6 returns true if the ip is ipv6
```
eurijon, line 388: typo: isIpv6

test/extended/openstack/utils.go
```diff
@@ -332,3 +384,29 @@ func conditionsMatchExpected(expected, actual map[string]string) bool {
...
+	netIP := net.ParseIP(ip)
+	if netIP != nil && netIP.To4() == nil {
+		ipv6 = true
+	} else {
+		ipv6 = false
+	}
+	return ipv6
+}
+
+// IsIPv4 returns true if the ip is ipv4
```
eurijon, line 401: typo: isIpv4

eurijon (commented):
Just mentioned two tiny typos, which I'm ok with if there is no any other change, everything else lgtm!

test/extended/openstack/utils.go
```diff
@@ -216,6 +217,57 @@ func getNetworkType(ctx context.Context, oc *exutil.CLI) (string, error) {
 	return networkType, nil
 }
 
+// Check the cluster networks and returns true if it finds one ipv4 and one ipv6 network there.
+func isDualStackCluster(ctx context.Context, oc *exutil.CLI) (bool, error) {
+	ipv4Found := false
+	ipv6Found := false
+
+	networks, err := oc.AdminConfigClient().ConfigV1().Networks().Get(ctx, "cluster", metav1.GetOptions{})
```
MaysaMacedo, line 225: Instead of fetching the network info twice, here and at line 255, it's better to only fetch it once and pass it as parameter.  Maybe at line 249.

test/extended/openstack/utils.go
```diff
@@ -332,3 +384,29 @@ func conditionsMatchExpected(expected, actual map[string]string) bool {
...
+}
+
+// isIpv4 returns true if the ip is ipv4
+func isIpv4(ip string) bool {
+	ipv4 := false
+
+	netIP := net.ParseIP(ip)
+	if netIP != nil && netIP.To4() != nil {
+		ipv4 = true
+	} else {
```
MaysaMacedo, line 408: this is not needed if the default value at line 403 is already false

test/extended/openstack/utils.go
```diff
@@ -332,3 +384,29 @@ func conditionsMatchExpected(expected, actual map[string]string) bool {
...
 }
+
+// isIpv6 returns true if the ip is ipv6
+func isIpv6(ip string) bool {
+	ipv6 := false
+
+	netIP := net.ParseIP(ip)
+	if netIP != nil && netIP.To4() == nil {
+		ipv6 = true
+	} else {
```
MaysaMacedo, line 395: same improvement proposed for the `isIpv4` function can be done here

test/extended/openstack/egressip.go
```diff
@@ -73,6 +73,12 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][egressip] An egre
 
 	g.It("attached to a floating IP should be kept after EgressIP node failover with OVN-Kubernetes NetworkType", func() {
 
+		dualstackIpv6Primary, err := isIpv6primaryDualStackCluster(ctx, oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if dualstackIpv6Primary { //This test is covering and scenario that has no sense with ipv6 as there is no FIP/VIP association.
```
MaysaMacedo, line 78: typo: a scenario

I'm not sure I follow why this test wouldn't be applicable for IPv6. Where there is no FIP/VIP?

test/extended/openstack/loadbalancers.go
```diff
@@ -132,18 +132,19 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
...
 
 				g.By("Checks from openstack perspective")
 				lb, err := octavialoadbalancers.Get(loadBalancerClient, loadBalancerId).Extract()
 				o.Expect(err).NotTo(o.HaveOccurred())
 				o.Expect(lb.Provider).Should(o.Equal(strings.ToLower(lbProviderUnderTest)), "Unexpected provider in the Openstack LoadBalancer")
-
-				fip, err := getFipbyFixedIP(networkClient, lb.VipAddress)
-				o.Expect(err).NotTo(o.HaveOccurred())
-				o.Expect(fip.FloatingIP).Should(o.Equal(svcIp), "Unexpected floatingIp in the Openstack LoadBalancer")
+				if isIpv4(lb.VipAddress) { // No FIP assignation on ipv6
```
MaysaMacedo, line 143: typo: assignment

MaysaMacedo (commented):
Nice work! I have a couple of questions and suggestions.

test/extended/openstack/egressip.go
rlobillo: This testcase is checking that when an egressIP moves to other worker, Openshift is not changing the neutron port so the connection between the egressIP and the FIP is not broken.

On IPV6, we don't have FIP so there is no need to test it in ipv6primary.

I notice that "VIP" is not precise, so I am changing the comment.

test/extended/openstack/loadbalancers.go
```diff
@@ -577,6 +584,11 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 			svcPort := int32(8082)
 			labels := map[string]string{"app": depName}
 			drop_sourcerange := "9.9.9.9/32"
+			ipv6PrimaryDualStack, err := isIpv6primaryDualStackCluster(ctx, oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if ipv6PrimaryDualStack {
+				drop_sourcerange = "2001:db8:2222:5555::/64"
```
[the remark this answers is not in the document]
rlobillo: Good point about renaming the variable, thanks!

we are using both 9.9.9.9/32 and 2001:db8:2222:5555::/64 should be as dummy CIDRs to confirm that the traffic is blocked. So I think it's ok to use that CIDR.

test/extended/openstack/loadbalancers.go
rlobillo: Done

test/extended/openstack/loadbalancers.go
```diff
@@ -686,7 +698,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 			o.Expect(listeners[0].ID).Should((o.Equal(listenerId)), "Listener ID for LB '%s' has changed", loadBalancerId)
 
 			// Check connectivity to the service
-			connNumber = 10
+			connNumber = 100
```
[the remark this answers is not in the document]
MaysaMacedo: ack, this might be considered a regression, WDYT?

MaysaMacedo:
/lgtm

we need to understand if the bump on the connection time might be a regression

test/extended/openstack/loadbalancers.go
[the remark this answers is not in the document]
rlobillo: Reverting it back to see how it behaves on CI

rlobillo:
I am ok reverting the bump and see how it behaves on the CI. I observed that on a virt setup and probably it's more stable in CI setups.

PR #147, opened 2024-01-03
OCPBUGS-23352: Fix machineset machine role test

mdbooth:
The test was asserting that all MachineSets must label their machines as workers, which is not valid in a variety of situations. For example, we document the creation of 'infra' nodes, but as this is user-controlled data there are no constraints on what this value might be. We remove this test as there is no useful assertion we can make here.

Instead we assert that all machines get the labels from their machineset, which is consistent with the existing machineset machine test.


pierreprinetti:
/approve
/lgtm
Thanks!

/cherry-pick release-4.14

PR #154, opened 2024-01-29
OCPBUGS-27969: Fix a case where two FDs are the same

itzikb-redhat:
The following test fails when there are two Failure domains are the same "[sig-installer][Suite:openshift/openstack] ControlPlane MachineSet ProviderSpec template is correctly applied to Machines"

### DIFF
--- a/test/extended/openstack/cpms.go
+++ b/test/extended/openstack/cpms.go
@@ -71,7 +71,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] ControlPlane Mach
 			VolumeType  string
 		}
 		var cpmsFDs []FailureDomain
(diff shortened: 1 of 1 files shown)


rlobillo:
What do you mean by "two FDs that are the same"? You mean that 2 masters have the same FD definition?

If yes, could you please elaborate what the fix is doing?

itzikb-redhat:
> What do you mean by "two FDs that are the same"? You mean that 2 masters have the same FD definition?
> 
> If yes, could you please elaborate what the fix is doing?

You are right, the message should be clearer It's 2 masters have the same FD.
Instead of using a map for the machines FDs and count the number of keys we use a slice and count the number of elements which should be the same as the ones in CPMS

PR #161, opened 2024-04-05
Improve load-balancer health monitor tests

MaysaMacedo:
When a cluster only has one worker node, the pods get scheduled
in this worker node by default. In the case of ETP:local test
the requested two Pods were scheduled in the same worker node,
consequently only one LB member had ACTIVE state due to health monitors
and the test expected for two ACTIVE members. This commit fixes the issue by
ensuring that the amount of worker nodes is taken into consideration
when defining the amount of Pod replicas.

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -356,8 +356,26 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 
 				g.By("Creating Openshift deployment")
 				labels := map[string]string{"app": "lb-etplocal-dep"}
-				replicas := int32(2)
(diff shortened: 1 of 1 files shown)


eurijon:
Hey @MaysaMacedo, shouldn't be two members if the svc is backed by two pods (even in one master node deployment)?

MaysaMacedo:
@rlobillo even if the cluster does not have only one worker node, it's possible that the two Pods gets scheduled in the same worker node. We need to add protection for this case, which is missing in this PR, so I will add hold until I fix it.

/hold

eurijon:
@MaysaMacedo ack, it makes sense yep, as the LB monitors the nodeports and not the pods

mandre (commented):
thanks!
/lgtm
/approve

PR #164, opened 2024-04-16
Use ci-operator.yaml to define a root image

pierreprinetti:
Define a root image in this repository, so that it can be changed without futher patches to openshift/release


mandre (commented):
/lgtm
/approve

PR #166, opened 2024-04-18
[OCPBUGS-32308] Use Set to check CPMS FDs Equals to machines FDs

itzikb-redhat:
Instead of using that the number of Failure domains in CPMS is equal to the number of failure domains in the machines we use Set (For example when we have only one Failure Domain in CPMS

### DIFF
--- a/test/extended/openstack/cpms.go
+++ b/test/extended/openstack/cpms.go
@@ -11,6 +11,7 @@ import (
 	"github.com/stretchr/objx"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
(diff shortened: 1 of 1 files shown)


rlobillo:
What's the difference between cpmsFDs and cpmsFdsSet? Do we really need to keep both? I think the first one can be removed.

itzikb-redhat:
> What's the difference between cpmsFDs and cpmsFdsSet? Do we really need to keep both? I think the first one can be removed.

cpmsFdsSet is a Set , in case we have two identical FDs in CPMS the set will be different than that of cpmsFDs

rlobillo:
> > What's the difference between cpmsFDs and cpmsFdsSet? Do we really need to keep both? I think the first one can be removed.
> 
> cpmsFdsSet is a Set , in case we have two identical FDs in CPMS the set will be different than that of cpmsFDs

As dicussed, please remove cpmsFDs

itzikb-redhat:
> > > What's the difference between cpmsFDs and cpmsFdsSet? Do we really need to keep both? I think the first one can be removed.
> > 
> > 
> > cpmsFdsSet is a Set , in case we have two identical FDs in CPMS the set will be different than that of cpmsFDs
> 
> As dicussed, please remove cpmsFDs

Done

pierreprinetti:
/retitle OCPBUGS-32308: Use Set to check CPMS FDs Equals to machines FDs

PR #167, opened 2024-04-21
Test NFS share when a PVC with manila provider is created

itzikb-redhat:
- Create a PVC and check that a manila share is created
- Create a deployment of 2 pods with a mounted directory on this share
- Check that the two pods are able to write and read to/from the share
- Move createTestDeployment to utils

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/openshift/openstack-test
 go 1.20
 
 require (
-	github.com/gophercloud/gophercloud v1.4.0
+	github.com/gophercloud/gophercloud v1.11.0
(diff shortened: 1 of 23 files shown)


rlobillo:
This test is not passing on CI (For 4.16). 

`err: 
fail [/home/stack/openstack-test/test/extended/openstack/volumes.go:204]: Expected
    : 2
to equal
    : 1`

I'm not merging this one as we observe it passing in 4.16 to avoid noise on CI results.

EmilienM:
@rlobillo please the next time make sure the test pass well in our CI as well, don't hesitate to reach out to know which jobs can be triggered.
Thanks

PR #169, opened 2024-04-30
Add cinder test using default storageClass

rlobillo:
This test uses the default cinder storageClass to create and use a PVC and covers https://issues.redhat.com/browse/OCPBUGS-30951 on multiAZ scenarios.

### DIFF
--- a/test/extended/openstack/utils.go
+++ b/test/extended/openstack/utils.go
@@ -197,7 +197,8 @@ func GetSharesFromName(ctx context.Context, client *gophercloud.ServiceClient, s
 	return shares, nil
 }
 
-func FindStorageClassByProvider(oc *exutil.CLI, provisioner string) *storagev1.StorageClass {
+// return storageClass with specific provisioner. If seek_default is true, it will look also for the default annotation.
+func FindStorageClassByProvider(oc *exutil.CLI, provisioner string, seek_default bool) *storagev1.StorageClass {
 	scList, err := oc.AdminKubeClient().StorageV1().StorageClasses().List(context.Background(), metav1.ListOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
@@ -208,7 +209,16 @@ func FindStorageClassByProvider(oc *exutil.CLI, provisioner string) *storagev1.S
 	}
 	for _, sc := range scList.Items {
 		if sc.Provisioner == provisioner {
-			return &sc
+			if seek_default {
+				val, ok := sc.GetAnnotations()["storageclass.kubernetes.io/is-default-class"]
+				// if we're explicitly looking for a default, the annotation must exist
+				if ok && val == "true" {
+					return &sc
+				}
+			} else {
+				// if we're not looking for a default, simply return the first result
+				return &sc
+			}
 		}
 	}
 	e2e.Logf("no default storage class found")
--- a/test/extended/openstack/volumes.go
+++ b/test/extended/openstack/volumes.go
@@ -62,7 +62,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 			}
 
 			g.By("Gather prometheus PVCs before resizing")
-			initial_pvcs, err := getMonitoringPvcs(ctx, dc)
+			initial_pvcs, err := GetPVCsFromNamespace(ctx, dc, "openshift-monitoring")
 			o.Expect(err).NotTo(o.HaveOccurred())
 
 			g.By("Gather Openstack cinder volumes for the PVCs before resizing")
@@ -128,7 +128,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 			g.By("Active wait checking that the status after resizing is expected (max 1 minute)")
 			o.Eventually(func() error {
 				e2e.Logf("Gather prometheus PVCs after resizing")
-				resized_pvcs, err := getMonitoringPvcs(ctx, dc)
+				resized_pvcs, err := GetPVCsFromNamespace(ctx, dc, "openshift-monitoring")
 				if err != nil {
 					return err
 				}
@@ -165,7 +165,7 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 						if !found && initvol.Name == rszvol.Name {
 							found = true
 							if initvol.Size+1 != rszvol.Size {
-								return fmt.Errorf("Unexpected size on resized volume: %d", initvol.Size+1)
+								return fmt.Errorf("unexpected size on resized volume: %d", initvol.Size+1)
 							}
 							if rszvol.Status != "in-use" {
 								return fmt.Errorf("cinder volume not in-use Status")
@@ -175,8 +175,8 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 
 						}
 					}
-					if found != true {
-						return fmt.Errorf("Pre-existing cinder volume %q is gone.", initvol.Name)
+					if !found {
+						return fmt.Errorf("pre-existing cinder volume %q is gone", initvol.Name)
 					}
 				}
 				return nil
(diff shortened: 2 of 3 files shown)


test/extended/openstack/volumes.go
```diff
@@ -280,6 +352,17 @@ func getMonitoringPvcs(ctx context.Context, dc dynamic.Interface) ([]objx.Map, e
 	return objects(objx.Map(obj.UnstructuredContent()).Get("items")), nil
 }
 
+// return list of PVCs defined in a given namespace
+func GetPvcsFromNamespace(ctx context.Context, dc dynamic.Interface, namespace string) ([]objx.Map, error) {
```
itzikb-redhat, line 356: What do you think of putting it in utils?

test/extended/openstack/volumes.go
```diff
@@ -241,6 +241,78 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
...
+			o.Expect(len(cinderVolumes)).To(o.Equal(1))
+			volumeID := cinderVolumes[0].ID
+			_, err = volumes.Get(ctx, volumeClient, volumeID).Extract()
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("Create a file in the pod's attached volume, restart it and read what was written")
+			_, err = oc.Run("exec").Args(pods.Items[0].Name, "--", "bash", "-c", fmt.Sprintf("echo %s > /data/1", fileContent)).Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("Delete pod so it get restarted")
```
itzikb-redhat, line 299: Delete pod so a new one is created

test/extended/openstack/volumes.go
```diff
@@ -241,6 +241,78 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
...
+			err = oc.Run("delete").Args("pod", pods.Items[0].Name).Execute()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			err = e2edeployment.WaitForDeploymentComplete(clientSet, deployment)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			_, err = exutil.WaitForPods(clientSet.CoreV1().Pods(ns),
+				exutil.ParseLabelsOrDie(""), exutil.CheckPodIsRunning, 1, 3*time.Minute)
```
itzikb-redhat, line 305: Although not crucial in this case I wonder if not to use CheckPodIsReady

test/extended/openstack/volumes.go
```diff
@@ -241,6 +241,78 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+			restarted_pods, err := oc.AdminKubeClient().CoreV1().Pods(ns).List(ctx, metav1.ListOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("Read pod and confirm data is persistent")
+			out, err := oc.Run("exec").Args(restarted_pods.Items[0].Name, "--", "bash", "-c", "cat /data/1").Output()
```
itzikb-redhat, line 311: It's not the first time we use it, maybe it's worth to add a function to verify a content

itzikb-redhat (commented):
Overall looks good! Some comments

test/extended/openstack/volumes.go
```diff
@@ -280,6 +352,17 @@ func getMonitoringPvcs(ctx context.Context, dc dynamic.Interface) ([]objx.Map, e
 	return objects(objx.Map(obj.UnstructuredContent()).Get("items")), nil
 }
```
stephenfin, line 353: This appears to be dead code now. Can we remove it?

test/extended/openstack/volumes.go
stephenfin, line 356: Any chance we could capitalise PVC?

```suggestion
func GetPVCsFromNamespace(ctx context.Context, dc dynamic.Interface, namespace string) ([]objx.Map, error) {
```

test/extended/openstack/volumes.go
```diff
@@ -241,6 +241,78 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenStack pla
 			o.Expect(err).NotTo(o.HaveOccurred())
 			o.Expect(out).To(o.Equal(fileContent))
 		})
+
+		g.It("should create a cinder volume when using cinder default storage class", func(ctx g.SpecContext) {
+
+			var err error
+
+			cinderSc := FindDefaultStorageClassByProvider(oc, "cinder.csi.openstack.org")
+			o.Expect(cinderSc).NotTo(o.BeNil(), "default cinder-csi storageClass not found.")
```
stephenfin, line 250: So we're saying that we can *always* expect Cinder CSI driver to be deployed, yes?

stephenfin (changes requested):
2 comments and a question inline but otherwise this looks pretty good to me.

test/extended/openstack/volumes.go
rlobillo: yes! afaik, shiftstack installation should create this SC with the default annotation pointing to cinder-csi provisioner.

test/extended/openstack/volumes.go
rlobillo: Sure! Done.

test/extended/openstack/volumes.go
rlobillo: Ack.  Thanks for noticing!

test/extended/openstack/volumes.go
rlobillo: I prefer to keep it explicit for sake of clarity

test/extended/openstack/volumes.go
rlobillo: Ack

test/extended/openstack/volumes.go
rlobillo: Ack

test/extended/openstack/volumes.go
rlobillo: For the moment it's only used in volumes.go so I think it's ok to keep it there.

test/extended/openstack/volumes.go
rlobillo: I refactored a bit the FindStorageClassByProvider function because I noticed that manila SC does not have the "is_default" annotation.

test/extended/openstack/utils.go
```diff
@@ -205,7 +206,14 @@ func FindStorageClassByProvider(oc *exutil.CLI, provisioner string) *storagev1.S
...
 		if sc.Provisioner == provisioner {
-			return &sc
+			if seek_default {
+				val, ok := sc.GetAnnotations()["storageclass.kubernetes.io/is-default-class"]
+				if ok && val == "true" {
+					return &sc
+				} else {
+					return &sc
+				}
+			}
```
stephenfin, line 216: As this stands, this will always fail with `no default storage class found` (or one of the earlier failures) unless `seek_default = true`. Is that expected? I had expected to see an else. How about something like this?

```go
if seek_default {
    val, ok := sc.GetAnnotations()["storageclass.kubernetes.io/is-default-class"]
    // if we're explicitly looking for a default, the annotation must exist
    if !ok {
        e2e.Failf("could not find annotation on storageclass")
    }
    if val == "true" {
        return &sc
    }
} else {
    // if we're not looking for a default, simply return the first result
    return &sc
}
```

stephenfin (commented):
I think the logic is slightly off, but I'd like to be sure

test/extended/openstack/utils.go
rlobillo: You are right @stephen. It is a typo in the indentation during copy/paste :/

I'll make use of your suggestion and test. Thank you!

stephenfin (approved):
This looks good to me now :+1:

PR #175, opened 2024-06-14
OCPBUGS-35505: LB tests: update sharding domain

EmilienM:
The domain has changed so let's update it and add a TODO so at some point we can figure out how to dynamically get this from the cluster.


EmilienM:
/hold
Like I said in the commit message, I haven't found yet the root cause of the change.

EmilienM:
I'll need a reproducer on this one :/

mandre (commented):
The code looks good, but it doesn't seem to have worked. The CI still fails on the same error:
```
Domain not found in the certificate on try #0
Expected
    <bool>: false
to be true
```

We'll need more debug.

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_openstack-test/175/pull-ci-openshift-openstack-test-main-test/1801703318293057536

PR #178, opened 2024-07-07
OCPBUGS-35064: Fetch a cluster external network for egress ip

MaysaMacedo:

### DIFF
--- a/test/extended/openstack/egressip.go
+++ b/test/extended/openstack/egressip.go
@@ -157,10 +157,17 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][egressip] An egre
 		g.By("Checking that the allowed_addresses_pairs are properly updated for the workers in the Openstack before failover")
 		checkAllowedAddressesPairs(ctx, networkClient, primaryWorker, secondaryWorker, egressIPAddrStr, machineNetworkID)
 
-		g.By("Creating a FIP in Openstack")
-		var fip *floatingips.FloatingIP
-		externalNetworkId, err := GetFloatingNetworkID(ctx, networkClient)
+		g.By("Fetching an external network for the cluster")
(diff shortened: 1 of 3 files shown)


test/extended/openstack/egressip.go
```diff
@@ -157,10 +157,17 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][egressip] An egre
...
-		g.By("Creating a FIP in Openstack")
+		g.By("Fetching an external network for the cluster")
+		cloudProviderConfig, err := getConfig(ctx,
+			oc.AdminKubeClient(),
+			"openshift-cloud-controller-manager",
+			"cloud-conf",
+			"cloud.conf")
+		o.Expect(err).NotTo(o.HaveOccurred())
 		var fip *floatingips.FloatingIP
-		externalNetworkId, err := GetFloatingNetworkID(ctx, networkClient)
```
mandre, line 162: is the `GetFloatingNetworkID()` function still used? Would it have made more sense to fix it so it's the one that makes the call to `GetClusterLoadBalancerSetting("floating-network-id", cloudProviderConfig)`?

mandre: OK, so looks like the only other use is in loadbalancer.go, where they're looking at the content of the clouds.conf first to see if there's a `floating-network-id` set, and otherwise fallback to calling `GetFloatingNetworkID()`.

Perhaps we should have `GetFloatingNetworkID()` be the one who tries to get the floating network id from the setting first, so we make the egress ip test more robust and can simplify the loadbalancer test a bit.

MaysaMacedo: good point! I just updated it.

MaysaMacedo:
@mandre @rlobillo this is ready for another look, when possible.

mandre (commented):
/lgtm
/approve

PR #181, opened 2024-07-29
Add skip for server group tests

eurijon:
The server group tests rely on the Machine object, as the server group is extracted from the '.spec.providerSpec.value.serverGroupName' specification of the machine.

This PR adds a skip for the server group tests when there are no Machine objects in the cluster (even if the machine-api operator is deployed), which is the case for UPI clusters.

### DIFF
--- a/test/extended/openstack/servergroup.go
+++ b/test/extended/openstack/servergroup.go
@@ -11,7 +11,9 @@ import (
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
 	"github.com/openshift/openstack-test/test/extended/openstack/client"
+	"github.com/openshift/openstack-test/test/extended/openstack/machines"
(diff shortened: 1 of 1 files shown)


pierreprinetti (approved):
LGTM! Thanks.

pierreprinetti:
This might warrant a backport. Do you mind opening a bug so that we can let automation cherrypick?

eurijon:
> This might warrant a backport. Do you mind opening a bug so that we can let automation cherrypick?
Let's see if the automation takes care of it.

/cherry-pick release-4.16

eurijon:
> > This might warrant a backport. Do you mind opening a bug so that we can let automation cherrypick?
> 
> Let's see if the automation takes care of it.
> 
> /cherry-pick release-4.16

Right, there was no bug...

eurijon:
Cherry-picking again after opening the bug

/cherry-pick release-4.16

PR #186, opened 2024-09-11
Remove Infra nodes from workers list

itzikb-redhat:
The following test create a deployment with number of pods equals to number of workers.
In a setup with Infra nodes this cause pods to be in pending state because Infra nodes are counted as workers. 
This patch removes the Infra nodes from the workers nodes list

```
"[sig-installer][Suite:openshift/openstack][lb][Serial] The Openstack platform should apply lb-method on TCP Amphora LoadBalancer when a TCP svc with monitors and ETP:Local is created on Openshift"
"[sig-installer][Suite:openshift/openstack][lb][Serial] The Openstack platform should apply lb-method on UDP Amphora LoadBalancer when a UDP svc with monitors and ETP:Local is created on Openshift"
```


mandre (commented):
/lgtm
/approve

PR #191, opened 2024-09-24
Adapt etcd-to-ephemeral test

rlobillo:
https://issues.redhat.com/browse/SOSQE-1875

It's observed that it's not needed to include the AdditionalBlockDevice in the CPMS object to have the ephemeral volume available to be mounted. That makes the test invalid. This patch is updating it so it properly test the feature.

### DIFF
--- a/test/extended/openstack/local-bd-etcd.go
+++ b/test/extended/openstack/local-bd-etcd.go
@@ -1,17 +1,17 @@
 package openstack
 
 import (
-	"fmt"
-	"strconv"
-
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack"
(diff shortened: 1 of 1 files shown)


EmilienM:
/test e2e-openstack-ccpmso-zone
This CI job should run this test.

EmilienM:
> What do you think?

I think that we should fix the job.
I won't block this PR from merging if that's what you need though.

rlobillo:
> > What do you think?
> 
> I think that we should fix the job. I won't block this PR from merging if that's what you need though.

Yep, the job should be fixed as a resolution of https://issues.redhat.com/browse/OCPBUGS-42587. I think @MiguelCarpio is on it.

By merging this PR, that job will be red but with a valid reason IMO.

PR #192, opened 2024-09-24
Fetch a cluster external network for egress ip

MaysaMacedo:

### DIFF
--- a/test/extended/openstack/utils.go
+++ b/test/extended/openstack/utils.go
@@ -472,7 +472,11 @@ func IsOctaviaVersionGreaterThanOrEqual(client *gophercloud.ServiceClient, const
 }
 
 // GetFloatingNetworkID returns a floating network ID.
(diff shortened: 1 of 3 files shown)


test/extended/openstack/utils.go
```diff
@@ -472,7 +472,11 @@ func IsOctaviaVersionGreaterThanOrEqual(client *gophercloud.ServiceClient, const
 }
 
 // GetFloatingNetworkID returns a floating network ID.
-func GetFloatingNetworkID(client *gophercloud.ServiceClient) (string, error) {
+func GetFloatingNetworkID(ctx context.Context, client *gophercloud.ServiceClient, cloudProviderConfig *ini.File) (string, error) {
```
mandre, line 475: Just a nit, `ctx` parameter is unused and we could drop it.

MaysaMacedo: done

mandre (commented):
/lgtm
/approve

PR #193, opened 2024-09-26
OCPBUGS-42498: Fetch a cluster external network for egress ip

MaysaMacedo:


mandre (commented):
This is a manual cherry-pick of https://github.com/openshift/openstack-test/pull/192
/lgtm

mandre (commented):
/approve

PR #194, opened 2024-10-17
OCPBUGS-43500: Change lbMethod check in dualstack LB tests

eurijon:
When the lbMethod configured in the cloud controller manager is SOURCE_IP but the service is not configured with ETP:Local, the algorithm that is finally applied to LB queries is round robin, as kubeproxy is applying it's own round robin algorithm in all the OCP nodes.

This change applies then the same replacement of the expected LB method as it's done in loadbalancer tests (in loadbalancers.go).

The tests

- sig-installerSuite_openshift/openstacklbSerialThe_Openstack_platform_should_create_a_TCP_Amphora_LoadBalancer_when_a_TCP_svc_with_type_LoadBalancer_and_IP_family_policy_PreferDualStack_and_IP_familiesIPv4_IPv6_is_created_on_Openshift
- sig-installerSuite_openshift/ openstacklbSerialThe_Openstack_platform_should_create_a_TCP_Amphora_LoadBalancer_when_a_TCP_svc_with_type_LoadBalancer_and_IP_family_policy_PreferDualStack_and_IP_familiesIPv6_IPv4_is_created_on_Openshift

were failing with SOURCE_IP LB method as it was expecting to receive answers from one service endpoint and it receives responses from two endpoints because the service is not configured with ETP:Local and kubeproxy is performing a round robin balancing (breaking the SOURCE_IP balancing).

More info about cloud controller manager options in https://docs.openshift.com/container-platform/4.17/installing/installing_openstack/installing-openstack-cloud-config-reference.html#ccm-config-lb-options

### DIFF
--- a/test/extended/openstack/dualstack.go
+++ b/test/extended/openstack/dualstack.go
@@ -34,6 +34,10 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 	availableLbProvidersUnderTests := [...]string{"Amphora", "OVN"}
 	ipFamilyPolicy := "PreferDualStack"
 	ipFamilies := [][]v1.IPFamily{{v1.IPv4Protocol, v1.IPv6Protocol}, {v1.IPv6Protocol, v1.IPv4Protocol}}
+	lbMethodsWithETPGlobal := map[string]string{
+		"OVN":     "source_ip_port",
+		"Amphora": "round_robin",
+	}
 
 	g.BeforeEach(func(ctx g.SpecContext) {
 		network, err := oc.AdminConfigClient().ConfigV1().Networks().Get(ctx, "cluster", metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


mandre (commented):
You need to run `go fmt` on the file to [make the verify check happy](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_openstack-test/194/pull-ci-openshift-openstack-test-main-verify/1846867004506509312#1:build-log.txt%3A32):
```
./hack/verify.sh
Files needing gofmt:
test/extended/openstack/dualstack.go
Run "go fmt ./..." 
```

EmilienM:
We have `e2e-openstack-dualstack` and `e2e-openstack-dualstack-v6primary` jobs, why weren't they triggered here?
We should use these jobs to run these tests if it's not the case already and also trigger them to confirm this patch did the work. Not just manual testing.

Thanks

eurijon:
> We have `e2e-openstack-dualstack` and `e2e-openstack-dualstack-v6primary` jobs, why weren't they triggered here? We should use these jobs to run these tests if it's not the case already and also trigger them to confirm this patch did the work. Not just manual testing.
> 
> Thanks

Thanks for the heads-up, noted for the next time. We need to consider adding the `pull-ci-*` openstack jobs to test the PRs automatically, which I think is not done yet.

PR #202, opened 2024-11-15
OSASINFRA-3535: Remove check for non-BlockStorage sections

stephenfin:
We are reworking the cinder-csi-operator to only copy the BlockStorage section and set everything else manually [1]. As such, we want to ignore all other sections. Make it so.

[1] https://github.com/openshift/csi-operator/pull/302


mandre (commented):
Where are you making use of skipSections?

EmilienM:
/testwith openshift/csi-operator/master/e2e-openstack openshift/csi-operator#302

mandre (commented):
/lgtm

PR #205, opened 2024-12-01
[OCPBUGS-45069] Add a check for manila share type

itzikb-redhat:
If the share type extra spec of driver_handles_share_servers is true we skip the test.

### DIFF
--- a/test/extended/openstack/utils.go
+++ b/test/extended/openstack/utils.go
@@ -18,6 +18,7 @@ import (
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/subnets"
 	"github.com/gophercloud/gophercloud/v2/openstack/sharedfilesystems/v2/shares"
+	"github.com/gophercloud/gophercloud/v2/openstack/sharedfilesystems/v2/sharetypes"
 	configv1 "github.com/openshift/api/config/v1"
 	machinev1 "github.com/openshift/api/machine/v1beta1"
 	operatorv1 "github.com/openshift/api/operator/v1"
@@ -195,6 +196,28 @@ func GetSharesFromName(ctx context.Context, client *gophercloud.ServiceClient, s
 		return emptyShare, err
 	}
 	return shares, nil
+}
+
+// return sharetype from openstack with specific name
+func GetShareTypesFromName(ctx context.Context, client *gophercloud.ServiceClient, shareTypeName string) ([]sharetypes.ShareType, error) {
+	var emptyShareTypes []sharetypes.ShareType
+	var allShareTypes []sharetypes.ShareType
+	var shareTypes []sharetypes.ShareType
+
+	allPages, err := sharetypes.List(client, sharetypes.ListOpts{}).AllPages(ctx)
+	if err != nil {
+		return emptyShareTypes, err
+	}
+	allShareTypes, err = sharetypes.ExtractShareTypes(allPages)
+	if err != nil {
+		return emptyShareTypes, err
+	}
+	for _, shareType := range allShareTypes {
+		if shareType.Name == shareTypeName {
+			shareTypes = append(shareTypes, shareType)
+		}
+	}
+	return shareTypes, nil
 }
 
 // return storageClass with specific provisioner. If seek_default is true, it will look also for the default annotation.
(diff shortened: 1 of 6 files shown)


rlobillo:
> If the share type extra spec of driver_handles_share_servers is true we skip the test.

Hi @itzikb-redhat.

Could you please elaborate why having `driver_handles_share_servers` set to `false` is a requirement? Do you know if that is documented somewhere? I took a look at the limitations and couldn't find it:

https://docs.openshift.com/container-platform/4.17/storage/container_storage_interface/persistent-storage-csi-manila.html#persistent-storage-csi-manila-limitations_persistent-storage-csi-manila

Maybe should be added there?

MiguelCarpio:
Hi @rlobillo @itzikb-redhat 

The test should validate if the `driver_handles_share_servers` is set to `true` or `false`, and in that way use the `share-network` or not.
* true --> Get the `share-network` (or create it and add it to the compute nodes) and use it to create the `manila-pvc`
* false --> Create the `manila-pvc` without a `share-network`

You can find in the following [documentation](https://docs.redhat.com/en/documentation/red_hat_openstack_platform/17.1/html/configuring_persistent_storage/assembly_manila-performing-operations-with-the-shared-file-systems-service_assembly-swift#ref_creating-nfs-cephfs-or-cifs-shares_performing-operations-with-manila) the requirements and what is supported to create a `manila-pvc` (or `share`):

<share_network>: The name of the share network:
* Required if the share type has driver_handles_share_servers set to true.
* Unsupported if the share type has driver_handles_share_servers set to false.
* Unsupported for CephFS-NFS and native CephFS. These protocols do not support share types that have driver_handles_share_servers set to true.

If this test is oriented to create a `manila-pvc` not using a `share-network`, it should skip when finding the `driver_handles_share_servers` is set to `true`.

rlobillo:
> Hi @rlobillo @itzikb-redhat
> 
> The test should validate if the `driver_handles_share_servers` is set to `true` or `false`, and in that way use the `share-network` or not.
> 
> * true --> Get the `share-network` (or create it and add it to the compute nodes) and use it to create the `manila-pvc`
> * false --> Create the `manila-pvc` without a `share-network`
> 

In our CI we are using below share type:
```
$ openstack share type show default
+----------------------+--------------------------------------+
| Field                | Value                                |
+----------------------+--------------------------------------+
| id                   | c0b38029-e7ab-48c3-bc2a-522237fcf89d |
| name                 | default                              |
| visibility           | public                               |
| is_default           | True                                 |
| required_extra_specs | driver_handles_share_servers : False |
| optional_extra_specs | snapshot_support : True              |
| description          | None                                 |
+----------------------+--------------------------------------+
```

and the test is passing. I believe that feature "driver_handles_share_servers" (also known as DHSS) is for instructing the backend to use different set of storage servers. We are not testing that AFAIK.

rlobillo:
I wonder if we should create a separate test then covering the `DHSS=true` case.

MiguelCarpio:
> > Hi @rlobillo @itzikb-redhat
> > The test should validate if the `driver_handles_share_servers` is set to `true` or `false`, and in that way use the `share-network` or not.
> > 
> > * true --> Get the `share-network` (or create it and add it to the compute nodes) and use it to create the `manila-pvc`
> > * false --> Create the `manila-pvc` without a `share-network`
> 
> Thanks @MiguelCarpio! I still have my concerns about this. In our CI we are using below share type:
> 
> ```
> $ openstack share type show default
> +----------------------+--------------------------------------+
> | Field                | Value                                |
> +----------------------+--------------------------------------+
> | id                   | c0b38029-e7ab-48c3-bc2a-522237fcf89d |
> | name                 | default                              |
> | visibility           | public                               |
> | is_default           | True                                 |
> | required_extra_specs | driver_handles_share_servers : False |
> | optional_extra_specs | snapshot_support : True              |
> | description          | None                                 |
> +----------------------+--------------------------------------+
> ```
> 
> and the test is passing. I believe that feature "driver_handles_share_servers" (also known as DHSS) is for instructing the backend to use different set of storage servers. We are not testing that AFAIK.

Yes, it passes because with `driver_handles_share_servers` set to `false` doesn't need a `share-network` to create a `manila-pvc`

rlobillo:
@itzikb-redhat I think it would be good to have a comment in the test itself with a link to the doc bug that explains this limitation.

/hold

EmilienM:
We should backport it.

PR #212, opened 2024-12-11
Check manila Storage class before checking manila endpoint

itzikb-redhat:
If manila storage class doesn't exist - skip the test. If it exists - check that the manila endpoint exists if not - fail


rlobillo:
I'm ok with this. The manilaSC existance is a consequence of some checks that the manila controller is doing. So if it exists and the endpoint is not there...it's a bug.

/lgtm

PR #215, opened 2024-12-17
[release 4.15] Check manila Storage class before checking manila endp…

itzikb-redhat:
Update the test "[sig-installer][Suite:openshift/openstack] The OpenStack platform on volume creation should create a manila share when using manila storage class "
If manila storage class doesn't exist - skip the test.
If it exists - check that the manila endpoint exists if not - fail


rlobillo:
@itzikb-redhat Could you please update the description?

itzikb-redhat:
> @itzikb-redhat Could you please update the description?

Done

PR #219, opened 2025-01-28
Skip noAllowedAddressPairs test when using external LB

MaysaMacedo:
When using external LB no allowed address pair is defined in the Port, consequently the test would fail as it expects for an allowed address pair to be set if it's is not on the Port created by the additional network added in the test.


EmilienM:
/approve
/lgtm
thanks

PR #222, opened 2025-02-03
OCPBUGS-50973: Fix noAllowedAddressPairs test

itzikb-redhat:
Fix for "[sig-installer][Suite:openshift/openstack] Bugfix ocpbug_1765: [Serial] noAllowedAddressPairs on one port should not affect other ports" test

The extra network was not added
If there was another network with noAllowedAddressPairs: true the test failed
because the port was supposed to have address pairs which was not true

* Modified machineNetworks from being a list of networks to a map where the value is whether
    the noAllowedAddressPairs is set to true, so if anther network has it it will be treated as it should (Having this field empty)
 * Modified Filter.ID of the extra network to UUID (Filter.ID is deprecated)
 * Moved json.Marshal to the right place

### DIFF
--- a/test/extended/openstack/ocpbug_1765.go
+++ b/test/extended/openstack/ocpbug_1765.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack"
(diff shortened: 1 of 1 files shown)


rlobillo:
@itzikb-redhat I don't get exactly the scenario you are fixing here. The original bug happened while adding a machine with several ports, so if one has the `noAllowedAddressPairs: false` then all of the ports for that machine will have it.

What you are doing in the test is replicating an existing machine and adding a port with `noAllowedAddressPairs: false`, correct? So the issue that this patch is fixing is when the original machine already has a port like that, right?

If yes, please elaborate what is the solution implemented here. Are you ignoring that ports in advance so the new machine does not include it? Are you marking them somehow so it won't be considered?

Thanks.

itzikb-redhat:
> @itzikb-redhat I don't get exactly the scenario you are fixing here. The original bug happened while adding a machine with several ports, so if one has the `1765.: false` then all of the ports for that machine will have it.
> 
> What you are doing in the test is replicating an existing machine and adding a port with `noAllowedAddressPairs: false`, correct? So the issue that this patch is fixing is when the original machine already has a port like that, right?
> 
> If yes, please elaborate what is the solution implemented here. Are you ignoring that ports in advance so the new machine does not include it? Are you marking them somehow so it won't be considered?
> 
> Thanks.

I'm adding a port with noAllowedAddressPairs: true. So in the bug if you set one port to true all of the ports would have the field empty as well. As for the fix, Since I'm building the machineset based on an existing worker machineset I can't rely on the fact that the extra port I'm adding will be the only one with the `allowed_address_pairs` empty which was my assumption before the fix. So I'm checking the machineset - if there is a network with `noAllowedAddressPairs: true` then `allowed_address_pairs` should be empty

mandre:
Doh! I forgot to refresh the page and thought it didn't perform the cherry-pick. All good then.

PR #223, opened 2025-02-08
Add observability test (for RHOSO only)

eurijon:
This PR adds the `[sig-installer][Suite:openshift/openstack] Creating ScrapeConfig in rhoso should trigger prometheus to add the rhoso target` test.

The test aims to verify the integration between shift-on-stack and Openstack telemetry systems, following the procedure described at [1]. It's only valid in clusters deployed on top of RHOSO.

It needs to be run with the next environment variables:
 - `OS_CLOUD`: openstack cloud name from clouds.yaml
 - `KUBECONFIG`: path to shift-on-stack cluster kubeconfig file
 - `RHOSO_KUBECONFIG`: path to Openstack management OCP cluster kubeconfig file
 - `SHIFTSTACK_PASS_FILE`: path to shift-on-stacl cluster kubeadmin password file

The test is skipped if any of `KUBECONFIG`, `RHOSO_KUBECONFIG` or `SHIFTSTACK_PASS_FILE` is not passed.

The test creates a scrapeconfig in Openstack so it can scrape some metrics from the shift-on-stack cluster.
It then checks the content of the `kube_node_info`, `kube_persistentvolume_info` and `cluster:master_nodes` metrics correlating the params between Openstack and shift-on-stack resources.

[1] https://github.com/openshift/installer/blob/main/docs/user/openstack/observability.md#method-b-scrape-ocp-metrics-from-rhoso


eurijon:
The test fails due to https://issues.redhat.com/browse/OSPRH-14025, will pass once fixed

test/extended/openstack/observability.go
```diff
@@ -0,0 +1,641 @@
...
+		openstackServerMap := make(map[string]string)
+		for _, server := range openstackServers {
+			e2e.Logf("Server found (ID: %v, Name: %v, Status: %v, HostID: %v)", server.ID, server.Name, server.Status, server.HostID)
+			openstackServerMap[server.Name] = server.ID
+		}
+		e2e.Logf("Openstack servers map: '%v'", openstackServerMap)
+
+		// Check every node obtained from metrics is in the servers info from Openstack
+		g.By("Checking every node obtained from metrics is in the servers info from Openstack")
+		for serverName, serverID := range nodeMetricsMap {
```
rlobillo, line 246: Instead of the for loop, have you tried with https://onsi.github.io/gomega/#containelementselement-any ?
Passing nodeMetricsMap and openstackServerMap can do the comparison for you. 

In your current code, I think it can be problematic if the metrics are providing duplicated entries, for example, right?

rlobillo:
@eurijon @itzikb-redhat Great work!

I suggest to split the test in three, one for checking the servers, one for the correlated metric and one for the cinder volume. The scrapeConfig setup can run at the beginning of all of them using the decorator BeforeAll inside the Ctx: https://onsi.github.io/ginkgo/#setup-in-ordered-containers-beforeall-and-afterall

I think this is the best approach because we can expect this kind of tests to grow in the coming releases, and adding more checks in separated testcases is helpful for all the stakeholders.

Thanks!

test/extended/openstack/observability.go
eurijon: After trying it:
```
o.Expect(openstackServerMap).Should(o.ContainElements(nodeMetricsMap),
    "The servers map obtained from Openstack '%v' does not contain the nodes obtained from metrics '%v'", openstackServerMap, nodeMetricsMap)
```
I've found out that `ContainElements` only works for checking values in a map, and not key-value pairs.
In addition the metrics map that is being generated:
```
map[ostest-xpwmx-master-0:d9f29887-14e8-43c2-90ef-ff3ac515d3cc ostest-xpwmx-master-1:6b2daf66-5871-42ea-9f63-7861daff01b8 ostest-xpwmx-master-2:74f4b0e7-d133-4627-8
2d7-16e2800332b5 ostest-xpwmx-worker-0-kctb6:889ee563-6cdf-407d-b900-a34c696b0f42 ostest-xpwmx-worker-0-zthwg:1bf4ae29-0e7f-4e63-9a68-6b1a5af1a6ea]
```
does not contain duplicated entries as the query 
```
group by (node, provider_id)  (
  kube_node_info
)
```
is filtering them, so that shouldn't be an issue.

eurijon:
> @eurijon @itzikb-redhat Great work!
> 
> I suggest to split the test in three, one for checking the servers, one for the correlated metric and one for the cinder volume. The scrapeConfig setup can run at the beginning of all of them using the decorator BeforeAll inside the Ctx: https://onsi.github.io/ginkgo/#setup-in-ordered-containers-beforeall-and-afterall
> 
> I think this is the best approach because we can expect this kind of tests to grow in the coming releases, and adding more checks in separated testcases is helpful for all the stakeholders.
> 
> Thanks!

Thanks for your review. I've tried with `BeforeAll` but I'm having lot of issues and I'm not able to make it run only once inside an ordered container. Looks like there are issues with `BeforeAll` and randomize and parallelism in openstack-tests.

I've splitted the test in three different tests and each one is now creating a different `scrapeConfig`. They can run in parallel (will do by default) so creating a `scrapeConfig` in each test (takes between 3~4 minutes) is not adding time overhead.

eurijon:
Tests passing:
```
passed: (3m43s) 2025-03-12T00:36:18 "[sig-installer][Suite:openshift/openstack] Check shift-on-stack kube_node_info metric in RHOSO prometheus"
passed: (3m43s) 2025-03-12T00:36:19 "[sig-installer][Suite:openshift/openstack] Check shift-on-stack correlated metric in RHOSO prometheus"
passed: (4m53s) 2025-03-12T00:37:28 "[sig-installer][Suite:openshift/openstack] Check shift-on-stack kube_persistentvolume_info metric in RHOSO prometheus"
```

test/extended/openstack/observability.go
rlobillo: great! Thanks

test/extended/openstack/observability.go
```diff
@@ -0,0 +1,797 @@
...
+		// Create a token in the shiftstack cluster
+		g.By("Creating a token in the shiftstack cluster")
+		SetTestContextHostFromKubeconfig(shiftstackKubeConfig)
+		pass, err := os.ReadFile(shiftstackPassFile)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		_, err = oc.Run("login").Args("-u", "kubeadmin").InputString(string(pass) + "\n").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		shiftstackToken, err = oc.Run("whoami").Args("-t").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		e2e.Logf("Token: '%s'", shiftstackToken)
```
mandre, line 112: We should probably avoid logging the token.

mandre (commented):
Just a nit, looks good otherwise.

test/extended/openstack/observability.go
eurijon: Sure, I've removed it

eurijon:
> I don't see the tests passing, in [the latest run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_openstack-test/223/pull-ci-openshift-openstack-test-main-test/1899610449045884928), the 3 tests are skipped:
> 
> ```
> Reason: skip [github.com/openshift/openstack-test/test/extended/openstack/observability.go:82]: KUBECONFIG, SHIFTSTACK_PASS_FILE and RHOSO_KUBECONFIG env vars must be set for this test to run Ginkgo exit error 3: exit with code 3
> ```
> 
> Am I missing something?
> 
> edit: replying to myself, I imagine these test ran downstream as we don't have a RHOSO environment we can run the tests again upstream.

Right, I got the successful results on a RHOSO deployment where `openstack-tests` is run with the `KUBECONFIG`, `RHOSO_KUBECONFIG` and `SHIFTSTACK_PASS_FILE` env vars, otherwise the tests are skipped as they are only applicable to RHOSO deployments.

PR #233, opened 2025-04-14
OSASINFRA-3754: CCM: Remove check for clouds.conf key

stephenfin:
The config file stored here was used for the legacy CCM. It is not relevant to the external CCM and we would like to stop setting it in https://github.com/openshift/installer/pull/9601. Remove the test, allowing us to do so.


mandre (commented):
/lgtm
/approve

stephenfin:
/verified

?

stephenfin:
/verified by "[sig-installer][Suite:openshift/openstack] The Openshift on cloud provider configuration should store cloud credentials on secrets", @stephenfin

PR #234, opened 2025-04-23
Update ConfigMap to check the LB configuration

rlobillo:
The new config map is present also in hypershift clusters, so we can use both for the same purpose: gathering the LB config in the OCP cluster.

### DIFF
--- a/test/extended/openstack/dualstack.go
+++ b/test/extended/openstack/dualstack.go
@@ -70,8 +70,8 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 		g.By("Gathering cloud-provider-config")
 		cloudProviderConfig, err = getConfig(ctx,
 			oc.AdminKubeClient(),
-			"openshift-cloud-controller-manager",
-			"cloud-conf",
+			"openshift-config",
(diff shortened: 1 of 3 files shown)


rlobillo:
/hold

Leaving this to @imatza-rh , as we need to use different configMap for hosted cluster 

@EmilienM do you think this is problematic?

rlobillo:
Testing another approach: 

			userConfigMap := configParams{
				namespace:    "openshift-config",
				name:         "cloud-provider-config",
				key:          "config",
				skipSections: nil,
				skipKeys:     nil,
				
Comparing both hypershift and regular OCP, both look updated.

rlobillo:
> Testing another approach:
> 
> ```
> 			namespace:    "openshift-config"
> 			name:         "cloud-provider-config"
> 			key:          "cloud.conf"
> ```
> 
> Comparing both hypershift and regular OCP, both look updated.

So it's not working because hosted cluster uses "cloud.conf" and regular cluster use "config". Furthermore, there are defaults params that are not present in that CM, so we should keep using the one before this patch for regular clusters. 

All points to create a function that will let the test to know if it's an hypershift cluster or not and set the CM to check accordingly.

PR #238, opened 2025-05-28
Add manual setting of enable_topology test

eurijon:
The test sets `enable_topology` to `false` in the cloud provider config and waits until it's reflected in the CSI driver config. It then creates a pvc and a deployment with a pod mounting it, checking that pvcs can be bounded in a cluster where `enable_topology` is `false` without the need of having a mismatch between compute and volume AZs.

The test runs serially as the cloud provider config is changed so the rest of the tests are not impacted, and the config is reverted once the test finishes.

The test is skipped if ´enable_topology´ is not `true` in the CSI driver config.

Note that `enable_topology` is auto-configured by the CSI driver controller, and this test is changing the value to check the compute and volume AZ mismatch scenario (something a user is allowed to do).

### DIFF
--- a/test/extended/openstack/topology.go
+++ b/test/extended/openstack/topology.go
@@ -1,18 +1,27 @@
 package openstack
 
 import (
+	"context"
+	"fmt"
+	"strconv"
+
 	"github.com/openshift/openstack-test/test/extended/openstack/client"
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack"
 	vavailabilityzones "github.com/gophercloud/gophercloud/v2/openstack/blockstorage/v3/availabilityzones"
+	"github.com/gophercloud/gophercloud/v2/openstack/blockstorage/v3/volumes"
 	cavailabilityzones "github.com/gophercloud/gophercloud/v2/openstack/compute/v2/availabilityzones"
 	g "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


test/extended/openstack/topology.go
```diff
@@ -81,4 +90,234 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack] The OpenShift clu
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+		err = e2edeployment.WaitForDeploymentComplete(clientSet, deployment)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		// Get the PVCs from the test namespace
+		pvcs, err := GetPVCsFromNamespace(ctx, dc, ns)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(len(pvcs)).To(o.Equal(1))
+
+		// Check the volume in openstack
```
itzikb-redhat, line 175: Should we have it in utils?

itzikb-redhat (commented):
Small comment , otherwise lgtm

stephenfin (commented):
/approve
/lgtm

Nice work here.

test/extended/openstack/topology.go
eurijon: Right, the functions related to volumes and PVS are defined in `volumes.go`. We can move them to `utils.go` but on the other hand we need to care about the size and structure of `utils.go` too.

PR #239, opened 2025-06-04
make: Check for updates in cmd before building

pierreprinetti:
Before this patch, changes in ./cmd/openshift-tests would not trigger recompilation when running `make`.


mandre (commented):
/lgtm

PR #240, opened 2025-06-05
Don't include infra nodes in worker node list

itzikb-redhat:
In server group tests.
Otherwise "[sig-installer][Suite:openshift/openstack] The OpenStack platform creates Worker nodes in a server group" test fails because infra nodes and worker nodes are not in the same group.


mandre (commented):
/lgtm
/approve

PR #243, opened 2025-06-11
[release-4.19] Add manual setting of enable_topology test

eurijon:
This is a manual cherry-pick of https://github.com/openshift/openstack-test/pull/238 + additional commit increasing the deployment creation waiting timeout.

The test sets enable_topology to false in the cloud provider config and waits until it's reflected in the CSI driver config. It then creates a pvc and a deployment with a pod mounting it, checking that pvcs can be bounded in a cluster with enable_topology is set to false without the need of having a mismatch between compute and volume AZs.

The test runs serially as the cloud provider config is changed so the rest of the tests are not impacted, and the config is reverted once the test finishes.

### DIFF
--- a/test/extended/openstack/topology.go
+++ b/test/extended/openstack/topology.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/openshift/openstack-test/test/extended/openstack/client"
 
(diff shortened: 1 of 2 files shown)


eurijon:
Closing this and reopening the automated cherry-pick https://github.com/openshift/openstack-test/pull/241 as it was an issue with the `openstack-cloud-provider` image not containing the latest commits
/close

PR #246, opened 2025-06-24
Disregard infra nodes in getFirstWorkerID

itzikb-redhat:
It was missed in https://github.com/openshift/openstack-test/pull/240" "[sig-installer][Suite:openshift/openstack] Bugfix ocpbug_1765: [Serial] noAllowedAddressPairs on one port should not affect other ports" is failed on a setup with Infra nodes


mandre (commented):
/lgtm
/approve

PR #251, opened 2026-03-24
NO-JIRA: Update OWNERS_ALIASES

pierreprinetti:
Remove pierreprinetti


mandre:
Done via https://github.com/openshift/openstack-test/pull/255

PR #252, opened 2026-04-27
OCPBUGS-84306: Remove certificate CN assertion from LB ingress test

mandre:
The IngressController LB test fails because the certificate CN served by the canary route through the shard IngressController does not match the hardcoded expected value.

The certificate controller generates a wildcard cert (`CN=*.<domain>`) for each IngressController, but the canary service mirrors the default IngressController's cert (`CN=*.apps.<cluster>`). Which cert the client sees depends on timing and TLS termination behavior, which has changed across OCP versions (see [1]).

Remove the certificate CN assertion entirely. The test's purpose is to verify that a TCP LoadBalancer is created when a LoadBalancerService IngressController is created, not to test certificate management. The test still validates the full TLS-encrypted path: connection, HTTP 200, and correct canary body across 100 iterations.

[1] https://github.com/openshift/openstack-test/pull/175

### DIFF
--- a/test/extended/openstack/loadbalancers.go
+++ b/test/extended/openstack/loadbalancers.go
@@ -573,11 +573,11 @@ var _ = g.Describe("[sig-installer][Suite:openshift/openstack][lb][Serial] The O
 			name := "shard-" + string(uuid.NewUUID())
 			domain := "ingress-canary.openshift-ingress-canary.svc"
 			shardIngressCtrl, err := deployLbIngressController(ctx, oc, 10*time.Minute, name, domain, map[string]string{"ingress.openshift.io/canary": "canary_controller"})
+			o.Expect(err).NotTo(o.HaveOccurred(), "new ingresscontroller did not rollout")
 			defer func() {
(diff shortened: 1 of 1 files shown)


mandre:
/hold

It didn't work:
```
Correct certificate CN was never observed after 100 attempts
```

mandre:
The job failed again, but at least now the `The Openstack platform should create a TCP Amphora LoadBalancer when LoadBalancerService ingressController is created on Openshift` test passes :green_circle:

stephenfin:
/approve
/lgtm

This is a simpler (and more correct, IMO) fix that the first revision.

mandre:
The issue with insight-operator has nothing to do with this PR, and it [affects platforms across the board](https://search.dptools.openshift.org/?search=all+containers+in+ns%2Fopenshift-insights+must+have+terminationMessagePolicy%3DFallbackToLogsOnError&maxAge=48h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job).
/override ci/prow/test

PR #260, opened 2026-05-19
OSASINFRA-4209: Migrate to openshift-tests-extension (OTE)

mandre:
Migrate to OTE using the `ote-migration` skill [1].

Add cmd/extension/main.go as a new OTE-based test extension binary
alongside the existing openshift-tests entry point. This registers
all openstack tests with the OTE framework, adds [OTP] annotations
to all Describe blocks, and integrates the extension build into the
Makefile and Containerfile.

The OTE binary discovers 51 tests and does not pull in origin's
monitor test framework, avoiding the duplicate test results seen
with the old entry point.

[1] https://github.com/openshift-eng/ai-helpers/tree/main/plugins/ote-migration

PR #261, opened 2026-05-21
Drop legacy entry point

mandre:
Depends on https://github.com/openshift/openstack-test/pull/260 and https://github.com/openshift/release/pull/79487.


mandre:
We need to switch _all_ the CI jobs, for all repo, all supported branches to use the new workflow before we can merge this patch.
/hold

PR #274, opened 2026-06-25
OSASINFRA-4373: Migrate to openshift-tests-extension (OTE), 4.15

danchild:
Migrate to OTE using the ote-migration skill [1].

- Mirror of PR #260 applied to release-4.15
- Original PR message and fix by @mandre

Add cmd/extension/main.go as a new OTE-based test extension binary
alongside the existing openshift-tests entry point. This registers
all openstack tests with the OTE framework, adds [OTP] annotations
to all Describe blocks, and integrates the extension build into the
Makefile and Containerfile.

The OTE binary discovers **44** tests and does not pull in origin's
monitor test framework, avoiding the duplicate test results seen
with the old entry point.

[1] https://github.com/openshift-eng/ai-helpers/tree/main/plugins/ote-migration

Previous OTE migrations used a monorepo approach; however, that
approach won't work for earlier openstack-test release versions.
The openstack-test entry point uses an older version
of ginkgo (**v2.6.1-0.20230317131656-c62d9de5a460**) whose API is incompatible with the 
openshift-tests-extension framework (**v2.6.1-0.20241205171354-8006f302fd12**). The incompatible versions
are a part of the same major 2 version release; therefore, a separate module is required to test OTE alongside
the existing openshift-tests. The module has the same form as that taken in previous migrations (e.g. https://github.com/openshift/openstack-test/pull/260). Both a **git mv** and changing line 25 on the Makefile are all that is needed to accomplish a unified structure across releases once the OTE tests are validated. Finally, the module's tests-extension is not vendored, because it is a temporary workaround and we don't want to include hundreds to thousands of objects that overlap with what is already vendored in the project's root module.


mandre (commented):
/lgtm
/approve

